PR #22, opened 2018-08-01
renaming cameras after ctapipe convention

vuillaut:
Hi.

I would like to use image-extractor as an *input* for ctapipe and the camera names are different from the convention chosen for ctapipe. I find it very confusing and don't see an obvious reason behind it so I'd like to propose to use the ctapipe convention for future versions of image-extractor.

I also removed unused packages and pycharm made some pep8 changes on the fly.

### DIFF
--- a/image-extractor/image_extractor.py
+++ b/image-extractor/image_extractor.py
@@ -2,12 +2,10 @@
 """
 Module for main ImageExtractor class and command line tool.
 """
-import pkg_resources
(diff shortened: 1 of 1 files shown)


nietootein:
Hi Thomas,

Thanks for this PR - it makes total sense to me to homogenize the notation with that of ctapipe. 

Best,

D.

Issue #26, opened 2018-09-14
Add necessary documentation


bryankim96:
In order to make the code more useful, the following documentation would be very useful:

- An up-to-date and detailed description of the current cta-ml format (all tables, their purposes, the datatypes of each column, etc.). This is the most critical.
- The version of ctapipe it is written for (as this is the key dependency and the most likely interface to change) This should be done after #15.
- Instructions for extending/modifying the code to handle new data formats (this will depend strongly on whether/when the rewrite in #25 occurs).
- Instructions on performance and mass processing. Given that this code is likely to be used to process large amounts of data, it would be useful to give instructions on how best to handle the procedure (multiple jobs/multithreading). Performance benchmarks could also be provided if available to give an idea of how long processing will take (per event).

@nietootein @aribrill Anything else you can think of?

bryankim96:
* An up-to-date and detailed description of the current cta-ml format (all tables, their purposes, the datatypes of each column, etc.). This is the most critical.

This page has now been updated for v0.7.2

* The version of ctapipe it is written for (as this is the key dependency and the most likely interface to change) This should be done after #15.

This is now done in https://github.com/cta-observatory/dl1-data-handler/commit/1a7f6851f897726793221a380547065af498434c.

TjarkMiener:
Sphinx documentation should be added.

PR #30, opened 2018-11-19
Implement Data Writer

### LINKED ISSUE #25: Refactoring to separate event/data loop from dumping to file
In addition to the bugfixes required by other issues and updating image-extractor to agree with the current version of ctapipe, it may be helpful to reorganize the code to better support different input data formats (i.e. VERITAS, FACT, etc.) and possibly different output data formats.

One possible structure is as follows:

```python

# Abstract class for loading events sequentially
# Event loaders for different input data formats to be implemented as subclasses
# For the ctapipe/simtel input pipeline this would be mostly a wrapper around an event_source
# with some additional processing
class EventLoader(ABC):
     def __init__(self, files):
        self.files = files
        self.num_events = # compute number of events (after cuts)

    def __iter__(self):
        return self

    def __next__(self):
         return self.get_next_event()

    @abstractmethod
    def get_next_event(self):
         # Implement event loading logic here
         # Should return some standardized format, like a dict or nested objects with attributes
         # what can be handled uniformly by DataDumper
         # probably selection cuts can be self-implemented here (apply a check before returning)
         # Any image processing should be implemented here, although if there are some common processing steps, those can possibly be delegated out to a separate class

     @abstractmethod
    def get_metadata(self):
         # Get dataset-related metadata (non-eventwise)

# Abstract class
# Defines output PyTables format
# Should implement outputting to multiple files, shuffling, splitting into train/test sets, any other output file formatting
# Only one implementation should be needed (for the cta-ml data format), but in theory others could be implemented
class DataDumper(ABC):
    def __init__(self, output_filenames, event_loader):
        self.output_filenames = output_filenames
        self.event_loader = event_loader

    @abstractmethod
    def dump(self):
         # Do all necessary PyTables file setup, formatting, etc.
         # Should be able to handle dumping to multiple files and possibly multithreading (?)
          for event in self.event_loader:
                # Process and dump
          # Do any post-processing like shuffling or splitting into train/test


```

This should make it possible to separately modify the event loading logic and the output data format. New data formats can be supported by implementing EventLoaders. This should make the code more maintainable and also hopefully reduce the need for hard-coding as well as make the data format clearer directly from the code.

bryankim96:
Resolves #25 and #26. 

The writer is based on the previous version of Image Extractor with the non-critical features stripped out and everything rewritten and refactored into new classes. 

The overall structure is based on a DL1DataWriter class that manages the entire writing pipeline (from multiple input files to multiple output files), and is internally based on reading with EventSources (from ctapipe) and writing with a new class which I named DL1DataDumper. With this design, it should be possible to support loading data from any format as long as it implements the interface specified by ctapipe for EventSources (ctapipe provides instructions on this in their documentation). It should also be possible to support writing to arbitrary output formats, as long as they implement the interface specified by DL1DataDumper. 

It is not ideal, in my mind, that we have to pass dictionaries of settings opaquely through DL1DataReader to reach the EventSources and DL1DataDumpers, but I was somewhat constrained in this because I wanted DL1DataDumper to be loosely similar to the structure of the existing ctapipe HDF5TableWriter, which is designed to operate on a single file and implement methods to write containers to that file. Because I followed this general design, it became impossible to instantiate and pass in a single DL1DataDumper to the DL1DataWriter when multiple output files are being written and instead a dictionary of settings needed to be passed.

The DL1DataWriter class uses the Python multiprocessing library to read/write files in parallel, which should make it more straightforward to process large numbers of files. Loading configuration files and runlists for processing are both done using PyYAML in the YAML format.

I've tested the code and have verified that it works on our data files. However it would be very good to get additional independent confirmation on this. I updated the README and tried to document the code, so if you have the chance, it would be great if you could try processing some files. This would also provide further confirmation that the multiprocessing part of the code is working robustly in different cases.

I still need to update the wiki page, as there were some minor changes to the naming of columns and other aspects of the data format which will break compatibility between the new file format and the old file format. Generally, in this new version I tried to align things as closely as possible with the ctapipe container naming conventions.

One unresolved issue is that many of the Monte Carlo header fields we want are still not yet implemented in the ctapipe container, so they are not being written.


README.md
```diff
@@ -126,24 +110,39 @@ If the package was installed with pip as described above, you can import and use
...
-#max events to read per file
-max_events = 50
-
-data_file = "/home/computer/user/data/simtel/test.simtel.gz"
-
-#create an ImageExtractor with default settings
-extractor = image_extractor.ImageExtractor(output_path,tel_type_list=['MST-SCT:SCTCam'])
-
-extractor.process_data(data_file,max_events)
+from dl1-data-handler import dl1-data-writer
```
MaxNoe: this is not valid python code (the `-` )

README.md
```diff
@@ -126,24 +110,39 @@ If the package was installed with pip as described above, you can import and use
...
+event_source_settings = {'setting1': 'value1'}
+
+data_dumper_class = MyDataDumperClass
+data_dumper_settings = {'setting2': 'value2'}
+
+def my_cut_function(event):
+    # custom cut logic here
+    return True
+
+data_writer = dl1-data-writer.DL1DataWriter(event_source_class=event_source_class,
```
MaxNoe: This is not valid python code `-`, it will be interpreted as `dl1` minus `data` minus `writer.DL...`

dl1-data-handler/dl1_data_writer.py
```diff
@@ -0,0 +1,641 @@
...
+
+                for name in table_names:
+                    try:
+                        table = self.file.get_node(name, classname='Table')
+                        table.cols._f_col(col_name).create_index()
+                    except:
+                        pass
+
+class DL1DataWriter:
+    """Class for managing data processing using event sources and DL1DataWriters.
```
aribrill: DL1DataWriters -> DLDataDumpers

dl1-data-handler/dl1_data_writer.py
```diff
@@ -0,0 +1,641 @@
...
+
+    Provides some options for controlling the output file sizes.
+
+    Attributes:
+        event_source_class (subclass of ctapipe.io.eventsource.EventSource):
+            A subclass of EventSource which will be used to load and yield events
+            as DataContainers.
+        event_source_class (dict): A dictionary of kwargs which will be passed
```
aribrill: event_source_class -> event_source_settings

dl1-data-handler/dl1_data_writer.py
```diff
@@ -0,0 +1,641 @@
...
+            into the constructor for the EventSource.
+        data_dumper_class (subclass of dl1_data_writer.DL1DataDumper):
+            A subclass of DL1DataDumper which will be used to write events from
+            the EventSource to output files.
+        data_dumper_class (dict): A dictionary of kwargs which will be passed
```
aribrill: data_dumper_class -> data_dumper_settings

dl1-data-handler/dl1_data_writer.py
```diff
@@ -0,0 +1,641 @@
...
+        output_file_size (int): Maximum size of each output file. If the total
+            amount of input data requested for a given output file exceeds this
+            size, the output will be split across multiple files.
+        events_per_file (int): Maximum number of events to write per output file.
+            If the total number of input events requested for a given output file
+            exceeds this number, the output will be split across multiple files.
+    """
+    def __init__(self,
+                event_source_class=None,
+                event_source_settings={},
```
aribrill: Better to use None as default argument instead of {}

dl1-data-handler/dl1_data_writer.py
```diff
@@ -0,0 +1,641 @@
...
+                data_dumper_class=CTAMLDataDumper,
+                data_dumper_settings={},
```
aribrill: Better to use None as default argument instead of {}

bryankim96:
That was my oversight, sorry! I just added those two fields and tested it. I had to update the ctapipe version from 0.6.0 to 0.6.1 as well, as x_max was only added to the event container in 0.6.1. 

Other things I also didn't do yet are updating the scripts in the scripts directory and adding unit tests. I think some of the scripts can be removed entirely, like prepare_cuts_dict.py and process_dataset.py, but plot_numtel.py could be reworked to be a general purpose script for plotting the distribution of events from a dataset in some parameter. visualize_events.py also seems like it would still be useful, although the rewrite should probably wait until the reader and processor are done. I think it also might be useful to add a script to auto-generate runlist files given some simple settings (it would be tedious to construct very long ones by hand). However, I think it would be best to make a 0.7.0 release first, once this PR is merged, and do those things in 0.7.1.

bryankim96:
Here's a sample file I generated locally:

[gamma.zip](https://github.com/cta-observatory/dl1-data-handler/files/2596668/gamma.zip)

vuillaut:
Hi @bryankim96 .
Sorry for the delay, I have been quite busy last weeks.
Thank you for the good work. I shall have a look at the code this week.

aribrill:
An issue to switching to ctapipe-style telescope types e.g. "MST-SCT:SCTCam" is that they aren't valid Python identifiers because of the special characters '-' and ':'. These means that the "<tel_type>_multiplicity" column can't be used directly with a PyTables condition string, and also that the multiplicity and indices columns can't be accessed using the table.cols.col syntax. I believe it's possible to evade this in table.where() by explicitly redefining the column name variables using the condvars mapping option, but this is somewhat ugly. To me, it would be better to choose valid column names in the first place.

I suggest using the tel_type format "<stripped_optics>_<stripped_camera>", with all alphanumeric characters stripped out of the optics and camera fields, e.g. "MSTSCT_SCTCam". Since we store the optics and camera fields in Array_Information it would be trivial to reconstruct the ctapipe tel_type, though we could store it as a separate column "ctapipe_tel_type" in Telescope_Type_Information also.

aribrill:
Also, the camera and optics correspond to a telescope type, not a specific telescope, so I think they should be stored as Telescope_Type_Information, not Array_Information.

vuillaut:
Hi.
Why do you care about the telescope type?
In ctapipe, telescopes are given by their camera type only:
"LSTCam", "NectarCam", ...

bryankim96:
> An issue to switching to ctapipe-style telescope types e.g. "MST-SCT:SCTCam" is that they aren't valid Python identifiers because of the special characters '-' and ':'. These means that the "<tel_type>_multiplicity" column can't be used directly with a PyTables condition string, and also that the multiplicity and indices columns can't be accessed using the table.cols.col syntax. I believe it's possible to evade this in table.where() by explicitly redefining the column name variables using the condvars mapping option, but this is somewhat ugly. To me, it would be better to choose valid column names in the first place.

> I suggest using the tel_type format "<stripped_optics>_<stripped_camera>", with all alphanumeric characters stripped out of the optics and camera fields, e.g. "MSTSCT_SCTCam". Since we store the optics and camera fields in Array_Information it would be trivial to reconstruct the ctapipe tel_type, though we could store it as a separate column "ctapipe_tel_type" in Telescope_Type_Information also.

Yes, I noticed this as well. I had to change the code because the same issue arises with attribute access on PyTables nodes. You have to use get_node() (or getattr()) because the camera type names are not valid identifiers(i.e. file.get_node('/SCT:SCTCam') to get that table, rather than file.MST-SCT:SCTCam). Your proposed solution sounds reasonable to me. However it is worth noting that the only field that this would cause problems for in practice seems to be the "[tel_type]_multiplicity" column, as selections can't be done on the "indices" columns anyways. It could be worth just making a special case there instead. With that said, if there are no objections I will go ahead and make the changes you suggest, as well as fix some formatting errors (my editor's linter wasn't working before).

Also, it might also be worth asking ctapipe about this/bringing it to their attention. There was some discussion of naming here: https://github.com/cta-observatory/ctapipe/issues/404 and I thought I read an issue thread on the topic of choosing the subarray/telescope/camera/optics string representations, but I can't seem to find it now. "[optics]:[camera]" was added as the string representation of TelescopeDescription over a year ago (https://github.com/cta-observatory/ctapipe/commit/25af01d8e73077d3e32365fee87ec12537646ab3#diff-12b7b15edc7704ef0590fe3a91aadfef) but it seems likely that they didn't consider that someone might want to use it as a Python identifier. To be fair, I can't see any use case where it would be necessary when working within the existing ctapipe containers, but maybe something similar might come up in the future.

bryankim96:
> Also, the camera and optics correspond to a telescope type, not a specific telescope, so I think they should be stored as Telescope_Type_Information, not Array_Information.

Good catch, I will fix this as well.

aribrill:
> Hi.
> Why do you care about the telescope type?
> In ctapipe, telescopes are given by their camera type only:
> "LSTCam", "NectarCam", ...

I'm looking at [this code](https://github.com/cta-observatory/ctapipe/blob/master/ctapipe/instrument/telescope.py), which defines telescopes as a combination of optics and camera type. It's true that image mapping is defined by camera type alone. One reason I can think of to include both the optics and camera in the telescope type is that the same camera could conceivably be used along with different optics in the same production, for a study of something like telescope harmonization or a mirror upgrade. Then those images should not be stored together.

bryankim96:
I think these changes should cover all of the points mentioned above. The code has been tested with default settings, but I haven't yet written a full set of unit tests. I will go ahead and update the wiki as well.

Here is a sample file with one event:

[gamma_20deg_0deg_runs26640-26648___cta-prod3-lapalma-2147m-LaPalma-SCT_cone10.zip](https://github.com/cta-observatory/dl1-data-handler/files/2666816/gamma_20deg_0deg_runs26640-26648___cta-prod3-lapalma-2147m-LaPalma-SCT_cone10.zip)

dl1_data_handler/dl1_data_writer.py
```diff
@@ -0,0 +1,798 @@
...
+import multiprocessing
+import warnings
+import sys
+
+import yaml
+import numpy as np
+import tables
+from ctapipe import io, calib
+
+import table_definitions as table_defs
```
vuillaut: Here you want to do `from . import table_definitions as table_defs` I think

vuillaut:
Hi @bryankim96 
When I run `python dl1_data_handler/dl1_data_writer.py example_runlist.yml`
With a diffuse gamma file from prod3b in the `example_runlist.yml`, the process starts and seems to work for a while then crashes:

```
[{'target': './output_file_1.h5', 'inputs': ['./gamma_40deg_0deg_run102___cta-prod3-lapalma3-2147m-LaPalma_cone10.simtel.gz']}]
I/O block extended by 302680 to 1302680 bytes
I/O block extended by 252556 to 1555236 bytes
gzip: error writing to output: Broken pipe
I/O block extended by 302680 to 1302680 bytes
I/O block extended by 252556 to 1555236 bytes
I/O block extended by 188416 to 1743652 bytes
I/O block extended by 6833712 to 8577364 bytes
I/O block extended by 1064960 to 9642324 bytes
I/O block extended by 1442552 to 11084876 bytes
I/O block extended by 1384448 to 12469324 bytes
I/O block extended by 1976552 to 14445876 bytes
Exception ignored in: <bound method CTAMLDataDumper.__del__ of <__main__.CTAMLDataDumper object at 0x1011a9438>>
Traceback (most recent call last):
  File "dl1_data_handler/dl1_data_writer.py", line 192, in __del__
    self.file.close()
  File "/Users/thomasvuillaume/anaconda/envs/cta-dev/lib/python3.5/site-packages/tables/file.py", line 2734, in close
    self._node_manager.shutdown()
  File "/Users/thomasvuillaume/anaconda/envs/cta-dev/lib/python3.5/site-packages/tables/file.py", line 574, in shutdown
    key, node = registry.popitem()
  File "/Users/thomasvuillaume/anaconda/envs/cta-dev/lib/python3.5/weakref.py", line 255, in popitem
    key, wr = self.data.popitem()
KeyError: popitem(): dictionary is empty
```

dl1_data_handler/dl1_data_writer.py
bryankim96: I think this change would have to be coupled with the other point you raised. If the script is going to be moved to a separate file, outside of the package, then the import should definitely be absolute (like it is now), not relative. If it stays where it is inside the package, I think either way would work fine.

bryankim96:
> > Hi.
> Ok sorry for my previous review, I misunderstood that now the script is in the dl1_data_handler dir.
> 
> But this is very confusing, why is the script in an installation directory?
> 
> Wouldn't be better to have it in the script folder and to import the installed package (and not to import locally as it's done now) ?

That's a good point. Generally, up until now I felt that it was fine to have dl1_data_writer.py serve both roles (both as a module containing all of the required classes and as an executable script itself). In particular this seemed ok because all of the classes and methods in dl1_data_writer.py are so closely tied to this one particular task (which is executed by the script) and probably aren't likely to be imported separately. However, given that we may try to integrate more closely with ctapipe, it would make sense to detach the script part (if I'm understanding your suggestion correctly).

> Hi @bryankim96
> When I run python dl1_data_handler/dl1_data_writer.py example_runlist.yml
> With a diffuse gamma file from prod3b in the example_runlist.yml, the process starts and seems to work for a while then crashes:

Hmmm... it looks like an exception was thrown (and ignored, as usual) in the CTAMLDataDumper destructor, my guess is that something went wrong in finalize(). I'm not sure exactly what the cause is, though, as I believe finalize was called successfully in my test runs.. Let me take a look at this.

bryankim96:
Ok, sounds good! I will make that change as well as track down this error that you saw. After the script exited, was there still an output file left? And if so, were there any problems with it?

vuillaut:
> After the script exited, was there still an output file left? And if so, were there any problems with it?

@bryankim96 
The file has been created, I am able to open it and browse through it. Hard to know if all the info there though...
The output file is 84MB for an input file of 905MB, is that a usual reduction factor you observed?

bryankim96:
I made the change you suggested as well as some other small improvements.

Regarding the error:

That's strange... it seems a bit on the small side compared to the compression ratios I've been seeing, which suggests that maybe something went wrong in the middle of processing (?). It's strange, though, as the error being thrown in the CTAMLDataDumper destructor made it seem like it probably reached the end of processing before crashing.

For reference, here are some rough values from processing a single data file (gamma):

1 input file, 1 output file, 1 thread (no parallelism)
Events: 600
Telescopes: 4 LST, 15 MST-SCT

Input file size: 948 MB
Output file size: 121 MB
Runtime: ~100 sec. (some variability depending on my computer's CPU load)

Compression factor: ~7.8 (with LZO compression level 1 + chunking + index on default columns)
Size/event: ~200 kB
Time/event: ~0.16 sec/event

These values are only very approximate, as there is probably some startup time associated with the script  (which will be amortized) as well as other things going on with multiple files (re-instantiating a DataLoader and DataWriter for each output) that will affect the timing. Ideally, I should've done the test with a larger number of files several times with timeit to average this stuff out, but this should be close. Also, the file sizes will obviously depend on what telescopes are in the subarray and the average trigger rate of the telescopes.

I'm still unsure as to what the cause of the bug you saw is... I can't reproduce it. I ran the following command with no errors:

```bash
python write_data.py --output_dir ~/sample_output/ ~/runlist.yml --debug
```

where the runlist looks the same as yours but with one of my local simtel files instead of yours.

My theory now is that it has something to do with the differences between our data files. The ones I tested on only have LST and MST-SCTs, while it seems from the filename that yours might be a larger subarray. Unfortunately, all the simtel files I have on hand are only for the LST/SCT subarray. There's nothing obvious in the code that I think should break it when adding more telescope types, but I think I need to run it on the same data file and reproduce the error to see what the problem is. Is there some way you could transfer the data file over or direct me to where I could get it (I think it's probably too large to upload via Github)?

vuillaut:
@bryankim96 
Thank you for taking my comments into consideration.
I forgot to say, but fantastic work overall on this refactoring.
Concerning the debug, let's talk offline to see how I can provide you with the same file.

Issue #35, opened 2019-02-28
Starting work on pre-processing and data reader


vuillaut:
Hi.
I just open a thread to discuss the developments to be done for the reading and pre-processing of the produced files.
As shown in a [ASWG conf call](https://indico.cta-observatory.org/event/2191/contributions/19084/attachments/14994/18886/cta_ml_vuillaume_20190228.pdf), I have a prototype reader for LST that I use with the standard analysis chain.
Also as discussed in Madrid, @mikael10j has done a lot of work for the pre-processing of images.

I think for a start, we should try to list the developments that we want to see done and order them by priorities.
I will start with:
- develop a reader for produced HDF5 files into ctapipe containers
- this reader should be set as a plug-in for ctapipe
- this has to be related to #31 from early stages to decide where we go

nietootein:
Thank you, Thomas, for opening this issue. For your convenience I'm pasting here Thomas' developments on github (please, correct me if the branch is wrong): 

https://github.com/vuillaut/cta-lstchain/tree/read_hdf5_dl1

aribrill:
I have started developing a [reader](https://github.com/cta-observatory/dl1-data-handler/blob/reader/dl1_data_handler/reader.py). The idea is for loading the data into a format usable by Tensorflow/PyTorch, as discussed in Madrid. The data loading is basically complete, pending some final testing. The image mapping is a placeholder intended to be replaced with the functions currently in CTLearn and the preprocessing is essentially a placeholder since I only implemented a few trivial functions.

This seems a bit different though. To clarify, you are discussing a reader of the DL1 files into a ctapipe container, correct? So there would be two different types of reader/data loader? However, they could share the same pre-processing and image mapping functions.

vuillaut:
> This seems a bit different though. To clarify, you are discussing a reader of the DL1 files into a ctapipe container, correct? So there would be two different types of reader/data loader? However, they could share the same pre-processing and image mapping functions.

You are right.
There should be different tasks or even project there:
- pre-processing
- data reader for Tensorlow/Pytorch
- data reader for ctapipe

mikael10j:
[...]

PR #36, opened 2019-03-08
Add DL1 reader for ML

aribrill:
This PR adds a reader, along with associated processor and image mapper, for reading the DL1 files produced by the writer into a form useful for machine learning applications (specifically Tensorflow and PyTorch). This is related to but distinct from the reader for ctapipe discussed in #35, though some components may be in common. 

The reader is fully-fledged and should be able to read data in mono, stereo, and multi-stereo modes. There is a caveat that event-wise intensity for stereo and multi-stereo modes is currently undefined and will return a NotImplementedError if called. The reader API is iteration:
```python
reader = DL1DataReader(file_list)
for example in reader:
    do_something(example)
```
The image mapper is a placeholder and is intended to be replaced by the [image mapping algorithms](https://github.com/ctlearn-project/ctlearn/blob/master/ctlearn/image_mapping.py) from CTLearn in a subsequent PR. The processor introduces the Transform class that provides a template for preprocessing functions, but only a few simple functions to demonstrate its use are currently implemented. More should be added in future PRs.

I tested the reader by loading two sample files (thanks @bryankim96) for gamma - Paranal and proton - La Parma, and verified they loaded correctly one at a time. I did not check loading two files at once, as these were from different arrays.


dl1_data_handler/processor.py
```diff
@@ -0,0 +1,73 @@
...
+            101: 0 # proton
+            }
+
+    def describe(self, description):
+        self.description = [
+            {**des, 'name': 'class_label'} for des
+            in description if des['name'] == 'particle_id']
+        return self.description
+
+    def transform(self, example):
```
bryankim96: Should transform() somehow implement a check that the example actually matches the description? It seems to me that in theory you could have an example with the same length as self.description but with the wrong order or different fields. The way things are set up now that wouldn't occur, but I feel like the Transform class is such a nice and useful design that I could imagine it being used in other settings, where it might be nice for each Transform to be more self-contained/robust.

dl1_data_handler/reader.py
```diff
@@ -0,0 +1,408 @@
...
+            # Select which telescopes from the full dataset to include in each
+            # event by a telescope type and an optional list of telescope ids.
+            selected_telescopes = {}
+            for tel_type in selected_tel_types:
+                available_tel_ids = telescopes[tel_type]
+                # Keep only the selected tel ids for the tel type
+                if tel_type in selected_telescope_ids:
+                    # Check all requested telescopes are available to select
+                    requested_tel_ids = selected_telescope_ids[tel_type]
+                    invalid_tel_ids = list(set(requested_tel_ids) -
```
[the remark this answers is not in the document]
aribrill: I removed the cast to list.

dl1_data_handler/image_mapper.py
```diff
@@ -0,0 +1,18 @@
...
+        self.image_shape = {
+                'LST_LSTCam': (110, 110, 1),
+                'MST_FlashCam': (112, 112, 1),
+                'MST_NectarCam': (110, 110, 1),
+                'SST1M_DigiCam': (96, 96, 1),
+                }
+   
+    """Returns: image (WHC np.float32 array)"""
+    def map_image(self, vector, tel_type):
+        return np.resize(vector, self.image_shape[tel_type])
```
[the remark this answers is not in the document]
aribrill: I added a comment to make it explicit that this method is just a placeholder. In general, the new shape won't have the same number of elements as the vector, so reshape wouldn't work. I chose reshape so it just copies the vector as needed to fill up an output array of the right size.

dl1_data_handler/processor.py
aribrill: This is a good idea. Maybe a good way to do this would be to add a validate option to the transform method that checks that the transformed example matches the description in terms of the number of elements and their types and shapes.

dl1_data_handler/image_mapper.py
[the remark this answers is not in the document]
bryankim96: Ok, that makes sense.

dl1_data_handler/image_mapper.py
```diff
@@ -0,0 +1,19 @@
+import numpy as np
+
+class ImageMapper():
+
+    def __init__(self, mapping_method, channels):#", settings=defaults, ..."):
+        self.mapping_method = mapping_method
+        self.channels = channels
+        self.mapping_tables = {}
+        self.image_shape = {
+            'LST_LSTCam': (110, 110, 1),
```
vuillaut: Why these shapes?

dl1_data_handler/processor.py
mikael10j: the Transform class indeed looks great. To go a bit further it could be a callable:
`class Transform():
    def __init__(self):
        self.description = []
    def describe(self, description):
        self.description = description
        return self.description
    def __call__(self, example):
        return example`
turning the process function of DL1DataProcessor into:
`def process(self, example):
        for transform in self.transforms:
            example = transform(example)
        return example`

dl1_data_handler/reader.py
```diff
@@ -0,0 +1,408 @@
...
+                  in f.root.Telescope_Type_Information.where(query)][0]
+        num_channels = len(self.image_channels)
+        vector = np.empty(shape=(length + 1, num_channels), dtype=np.float32)
+        # An "empty" pixel at index 0 is used to fill blank areas in image
+        vector[0, :] = 0.0
+        # If the telescope didn't trigger, the image index is 0 and a blank
+        # image of all zeros with be loaded
+        for i, channel in enumerate(self.image_channels):
+            vector[1:, i] = record[channel]
+        image = self.image_mapper.map_image(vector, tel_type)
```
mikael10j: Applying the image mapper in the _get_image function, and thus in the __getitem__ function might slow down the training. Ideally, it would be applied once, when creating the dataset. But it that case, the data need to be stored in RAM and this may not be possible when dealing with the whole array. What do you think ?

dl1_data_handler/image_mapper.py
TjarkMiener: These shapes are the default ones and correspond to oversampling. E.g. if you rotate the pixel positions for the LSTCam so that they are aligned with the x- and y-axis, you will end up having 55 "positions" on one axis, which will be expand by the oversampling method by a factor of 2. With 55 positions (later it will declare as "ticks" in the code) I mean the number of elements (for one dimension) after applying np.unique() to the 2D pixel position array. Especially for the LSTCam the other axis has less than 110 ticks. To square the image, we fill extra rows or columns (depending on the axis) with artificial zeros. This will result in the typical stretch that the oversampling method is introducing to the images. With the new rebinning and interpolation methods we will get rid of this stretch. However, the image_mapper have an argument, where you can set an arbitrary image shape (only for the new methods), which will overwrite the default values above. Once this PR is merged, I will transfer the image mapping to DL1DH.

aribrill:
> Trying to create a reader with 'mc_energy' event_info I get an error:
> 
> ```python
> In [1]: from dl1_data_handler.reader import DL1DataReader
> 
> In [2]: f = ["../projets_CTA/UCM/data_handler/gamma-diffuse_20deg_0deg_run1000___cta-prod3
>    ...: -demo-2147m-LaPalma-baseline-mono.h5"]
> 
> In [3]: d = DL1DataReader(f, event_info=["mc_energy"])
> ---------------------------------------------------------------------------
> AttributeError                            Traceback (most recent call last)
> <ipython-input-3-937e4c70f268> in <module>()
> ----> 1 d = DL1DataReader(f, event_info=["mc_energy"])
> 
> ~/dl1-data-handler/dl1_data_handler/reader.py in __init__(self, file_list, mode, selected_telescope_type, selected_telescope_ids, selection_string, intensity_selection, shuffle, seed, image_channels, mapping_method, mapping_settings, array_info, event_info, transforms)
>     270                     'tel_type': None,
>     271                     'base_name': None,
> --> 272                     'shape': f.root.Events._f_col(col).shape,
>     273                     'dtype': f.root.Events._f_col(col).dtype
>     274                     }
> 
> AttributeError: 'Table' object has no attribute '_f_col'
> 
> In [8]: tables.__version__
> Out[8]: '3.4.4'
> ```

Fixed in fa046ef57ef99bdddf552543d81a73ffd0c7a834.

dl1_data_handler/processor.py
aribrill: I turned Transform into a callable in b960247 and added a validation method in ce95dbc. Note that validation is performed at processing, not initialization, time. This is unavoidable because validation requires comparison with a specific example. I envision the workflow being turning on validation while testing the network and turning it off for final production to optimize speed (though the difference is likely small).

dl1_data_handler/reader.py
aribrill: That's a fair point, and yes, the idea here is to avoid having to load the entire dataset in RAM (or store a preprocessed dataset on disk). To minimize the effect of image mapping on training speed, @TjarkMiener (re)implemented all the CTLearn image mapping methods to operate as matrix transformations, optimizing performance at mapping time. The mapping matrices for the method(s) used are calculated at initialization. That way, the expensive computations (the slowest being cubic interpolation at ~30s for all telescope types) are only done once, and the map_image function is relatively fast.

PR #45, opened 2019-04-26
Added option to save non-triggered MC events.

bryankim96:
Fixed num_showers header parameter to correctly sum rather than
overwrite.
Moved all list and dict default arguments into constructor to avoid
potential issues with mutable default arguments.
Added validation checks for all file-level parameters.

### DIFF
--- a/dl1_data_handler/table_definitions.py
+++ b/dl1_data_handler/table_definitions.py
@@ -6,7 +6,7 @@ the output PyTables .h5 files.
 """
 
 from tables import (IsDescription, UInt32Col, UInt16Col, UInt8Col,
-                    Float32Col, StringCol, BoolCol)
+                    Float32Col, StringCol)
 
 
 class EventTableRow(IsDescription):
--- a/dl1_data_handler/table_definitions.py
+++ b/dl1_data_handler/table_definitions.py
@@ -6,7 +6,7 @@ the output PyTables .h5 files.
 """
 
 from tables import (IsDescription, UInt32Col, UInt16Col, UInt8Col,
-                    Float32Col, StringCol)
+                    Float32Col, StringCol, BoolCol)
 
 
 class EventTableRow(IsDescription):
@@ -23,6 +23,8 @@ class EventTableRow(IsDescription):
     obs_id : tables.UInt32Col
         Shower observation (run) id. Replaces old "run_id" in ctapipe r0
         container.
+    triggered : tables.BoolCol
+        Whether the (Monte Carlo) shower triggered the array
     shower_primary_id : tables.UInt8Col
         Particle type id for the shower primary particle. From Monte Carlo
         simulation parameters.
@@ -47,6 +49,7 @@ class EventTableRow(IsDescription):
 
     event_id = UInt32Col()
     obs_id = UInt32Col()
+    triggered = BoolCol()
     shower_primary_id = UInt8Col()
     core_x = Float32Col()
     core_y = Float32Col()
--- a/dl1_data_handler/table_definitions.py
+++ b/dl1_data_handler/table_definitions.py
@@ -23,8 +23,6 @@ class EventTableRow(IsDescription):
     obs_id : tables.UInt32Col
         Shower observation (run) id. Replaces old "run_id" in ctapipe r0
         container.
-    triggered : tables.BoolCol
-        Whether the (Monte Carlo) shower triggered the array
     shower_primary_id : tables.UInt8Col
         Particle type id for the shower primary particle. From Monte Carlo
         simulation parameters.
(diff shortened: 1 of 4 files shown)


aribrill:
What changes have been made to the file format (if any) with the save_nontriggered_events enabled and disabled? In most use cases only the triggered events would be used, so it would be convenient if the reader could easily select them.

Am I correct that the event table would first list all triggered, then all non-triggered events, as opposed to all events in strict order of event id?

bryankim96:
No changes to the file format, although I can add an additional column with a boolean flag indicating the whether the event has triggered. 

Yes that's right. It's unfortunate, but there is no combined iterator that yields both MC non-triggered and triggered events in eventid order. SimTelFile.iter_mc_events() yields all events, triggered and non-triggered, but it doesn't yield the image information even when it's available (only iter_array_events() does that). The ideal thing would be an iterator that yields all mc events and, when images were taken, also includes those. As it is, that is not implemented and the only way to get it would be to modify SimTelFile, which I do not want to do. So I had to do an extra loop over iter_mc_events() after getting all triggered events from SimTelEventSource and simply exclude the duplicates.

In practice it shouldn't be important as I can't imagine a scenario where you would need to have the data strictly ordered by event id (as we usually shuffle anyways).

bryankim96:
@nietootein Was worried about the processing speed/time when writing the extra nontriggered events. In his tests he noticed it was taking a huge amount of time to process/write the additional events. I investigated it by doing some in-depth profiling and it turns out he was absolutely right that this is a major issue. After profiling, I found a difference in runtime on the same set of files with and without non-triggered events to be about a factor of 40.

The good news is that I caught several things and was able to make improvements that achieved a ~12% decrease in runtime when writing only triggered events and an ~90% (!) decrease in runtime when writing triggered and nontriggered events. So the existing triggered event writing loop was slightly suboptimal and the nontriggered event writing significantly so.

The bad news is that even after doing everything I could, there's still a roughly a factor of ~4 increase in runtime when adding non-triggered events. This is mostly irreducible, as the biggest time consumers left are the actual reading/IO from the simtel file through SimTelEventSource/SimTelFile as well as some stuff that seems unavoidable to me (like creating lists of 0s for the index columns and getting the event row to write to). Because the number of events increases so dramatically when we include non-triggered events (you can see the numbers yourself in the summary, it's about a factor of ~5000 for the files I tested on), even when mostly doing things efficiently the runtime still increases a lot.

Given this, I guess we need to consider whether it's worth including this info when we resume processing. With the new changes, it looks like adding non-triggered events should increase the processing time by a factor of ~4 (or maybe a little less). Maybe there is some other alternative to writing the data into the h5 files that we could consider? Or at least it might be nice to do it separately/ in parallel so it doesn't hold up our training while we wait.

I've attached a text file summarizing the before/after results (4 profiling runs) as well as the profiling reports for all 4 runs. You can take a look to verify/get an idea of where the improvement was achieved.

[profiling notes.txt](https://github.com/cta-observatory/dl1-data-handler/files/3123478/profiling.notes.txt)
[test_1.txt](https://github.com/cta-observatory/dl1-data-handler/files/3123479/test_1.txt)
[test_2.txt](https://github.com/cta-observatory/dl1-data-handler/files/3123480/test_2.txt)
[test_3.txt](https://github.com/cta-observatory/dl1-data-handler/files/3123481/test_3.txt)
[test_4.txt](https://github.com/cta-observatory/dl1-data-handler/files/3123482/test_4.txt)

@aribrill I also added the flag you requested! Also one clarification I forgot to mention about the event order: in general, like you said, for each input file the events will be written in the order (triggered -> nontriggered), however when multiple input files are written to a single output file (which is usually the case), this pattern will be repeated for each input file. So it'll look like (input file 1 triggered, input file 1 nontriggered, input file 2 triggered, input file 2 nontriggered, etc.). This could be further complicated if you set an output file event/size limit, which will split the events across multiple output files. One important thing to note here is that it probably isn't a good idea to use the event count limit on output files when you're saving triggered events, as the event counts get extremely large and only a small fraction of them are actual triggered events, so you could end up with many files which don't have any triggered events. The same is true to a lesser extent with the output file size limit, but it's not as extreme, as the nontriggered events don't take up such a disproportionate fraction of the output files when considered in terms of size (because they don't include images).

Finally, one other minor thing to note is that currently the ctapipe version isn't being written correctly for some reason... pkg_resources can't get the version number. I haven't been able to track down why.

dl1_data_handler/table_definitions.py
```diff
@@ -6,7 +6,7 @@
 """
 
 from tables import (IsDescription, UInt32Col, UInt16Col, UInt8Col,
-                    Float32Col, StringCol)
+                    Float32Col, StringCol, BoolCol)
```
aribrill: I suppose BoolCol is no longer needed.

PR #46, opened 2019-05-03
CTLearn integration

aribrill:
This PR makes minor API changes for better compatibility with CTLearn, improves the num_examples method, and fixes bugs discovered during the integration process. It is required for ctlearn-project/ctlearn/pull/117.

### DIFF
--- a/dl1_data_handler/image_mapper.py
+++ b/dl1_data_handler/image_mapper.py
@@ -2,9 +2,9 @@ import numpy as np
 
 class ImageMapper():
 
-    def __init__(self, mapping_method, channels):#", settings=defaults, ..."):
-        self.mapping_method = mapping_method
-        self.channels = channels
+    def __init__(self, **kwargs):
+        self.mapping_method = kwargs['mapping_method']
+        self.channels = kwargs['channels']
         self.mapping_tables = {}
         self.image_shape = {
             'LST_LSTCam': (110, 110, 1),
--- a/dl1_data_handler/image_mapper.py
+++ b/dl1_data_handler/image_mapper.py
@@ -3,7 +3,7 @@ import numpy as np
 class ImageMapper():
 
     def __init__(self, **kwargs):
-        self.mapping_method = kwargs['mapping_method']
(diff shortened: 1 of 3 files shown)


dl1_data_handler/reader.py
```diff
@@ -168,7 +168,7 @@ def __init__(self,
         if image_channels is None:
             image_channels = ['charge']
         self.image_channels = image_channels
-        self.image_mapper = ImageMapper(mapping_method, image_channels,
+        self.image_mapper = ImageMapper(channels=self.image_channels,
```
mikael10j: Raises a KeyError if mapping_settings is None (cause defined as an empty dict)

dl1_data_handler/image_mapper.py
```diff
@@ -2,9 +2,9 @@
 
 class ImageMapper():
 
-    def __init__(self, mapping_method, channels):#", settings=defaults, ..."):
-        self.mapping_method = mapping_method
-        self.channels = channels
+    def __init__(self, **kwargs):
+        self.mapping_method = kwargs['mapping_method']
```
mikael10j: We could add a try to avoid the KeyError.
Can we also return the input vector of map_image if no mapping_method is provided ?

dl1_data_handler/reader.py
aribrill: The mapping_method is still needed, but I moved it to the mapping_settings dictionary since there's no reason it has to be separate.

dl1_data_handler/image_mapper.py
aribrill: I changed it to dict.get() with a default value of "vector".

aribrill: > Can we also return the input vector of map_image if no mapping_method is provided ?

I think this is a great suggestion. There could be a vector mapping method that does this and is the default.

aribrill:
I think the current behavior of the placeholder image_mapper is better because it returns square arrays that are suitable with use with CNNs, so the code can be tested (even though the results are nonsense).

Issue #49, opened 2019-05-03
Add image mapping method to return unmapped vectors


aribrill:
As mentioned by @mikael10j in #46, the image mapper should have a method to return the unmapped vectors. Because this behavior most closely reflects the HDF5 data, it should be the default method if no method is specified.

TjarkMiener:
Do we want to use the unmapped vectors? Why we don't use 'oversampling' as the default method?

aribrill:
I don't think we have any plans to use that in CTLearn, but a basic use case might be as input for a traditional NN or other ML algorithm expecting an input vector. An argument for returning vectors as default is that it represents the least amount of processing possible. An argument against would be that the images were originally 2D and the user expectation would be to return them that way, since the 1D vectors are an artifact of the HDF5 file format. @mikael10j any input on this?

mikael10j:
I totally agree with "the least amount of processing possible" argument.

aribrill:
I mean that the images come from a set of camera pixels with a particular 2D physical arrangement.

vuillaut:
> An argument against would be that the images were originally 2D and the user expectation would be to return them that way, since the 1D vectors are an artifact of the HDF5 file format. @mikael10j any input on this?

Hi @aribrill 
What do you mean the images were originally 2D?
In ctapipe they are certainly not.
Either we talk about waveforms, and they are 3D of shape `(n_gain, n_pixels, n_samples)` or we talk about calibrated images and their shape is `(n_gain, n_pixels)` and will most probably be `(n_pixels,)` since real data will contain only one channel (so 1D).

The format on disk does not really matter here, it's how the reader reads the data that makes the object (and thus the dimension).

vuillaut:
> I mean that the images come from a set of camera pixels with a particular 2D physical arrangement.

Ok but since they don't follow a grid arrangement, I don't really expect to get an image as a 2D array 🤔

aribrill:
> Either we talk about waveforms, and they are 3D of shape (n_gain, n_pixels, n_samples) or we talk about calibrated images and their shape is (n_gain, n_pixels) and will most probably be (n_pixels,) since real data will contain only one channel (so 1D).

Right. I spoke imprecisely and neglected the channel dimension. I was referring to two spatial coordinates.

> Ok but since pixels don't follow a grid arrangement, I don't really expect to get an image as a 2D array

You would if you worked on the SCT :smile: (but yes, that makes sense).

[...]

PR #50, opened 2019-05-07
Shift CTLearn imagemapper methods to DL1DH

TjarkMiener:
This PR is about the integration of the [CTLearn](https://github.com/ctlearn-project/ctlearn) image mapping methods into DL1DH. These image mapping methods are converting a 1D array into a 2D image, which consists of a square pixel grid. We implemented five different conversion methods and two methods, where we are shifting the camera pixels (only for cameras with hexagonal layout) to arrange a square grid. The structure of the implementation are the same for all methods! We perform all costly calculations of the methods beforehand and independent of the event images. The results are stored in mapping tables, which contains the assignment information. During the training process, the mapping tables are applied to each event images, which ensure us to preserve the fast runtime. As a result, we got almost the same runtime for all conversion methods.

Conversion methods:
- **Oversampling** converts each hexagonal pixel to four square pixels and reweight the charge to realign to a square grid in order to preserve the intensity.
- **Nearest interpolation** assign the pixel charge to the nearest neighbor of the square grid points.
- **Rebinning** interpret the intensity as a histogram and rebin the hexagonal histogram into a square one. Since our implementation approximately rebin the histogram, the intensity is only preserved to a certain point. 
Screenshot 2019-05-07 at 03 08 18 Credits: [Holch](https://arxiv.org/pdf/1711.06298.pdf) 
We are currently calculating the areas like this: We put 100 points per square pixels and then perform NN to determine in which hexagon the points are located. Improvements or/and ideas to calculate the areas analytically are welcome!
- For **bilinear** and **bicubic interpolation**, we add artificial pixels with zero intensity outside of the camera to draw Delaunay triangulation on the hexagonal grid. Then we perform bilinear and bicubic interpolation and norm the mapping tables, which approximately preserve the intensity.

All conversion methods are also available for the square pixel cameras. A requirement of these methods is the alignment of the camera pixel positions. So for LSTCam, NectarCam and MAGICCam we have to rotate the pixel position before applying the algorithm. We added the option 'rotate_camera' in the image_mapper, so that we are able rotate back the mapping tables to get the original shower orientation. However, an extra interpolation ([scipy.ndimage.rotate](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.rotate.html)) is needed for this step. There are two other important option in the image_mapper. The first one is interpolation_image_shape, which specifies the shape of the output grid. The second one is mask_interpolation. This will set all output grid points to zero, if there are located outside of the camera. If you only want to consider image charges in your analysis, the default 'channels' option should be selected. 

The image below shows the impact of the different conversion methods on a simulated proton event captured by *FlashCam*.

FlashCam_conversion_methods

Some preliminary comparing results of the different methods with the old CTLearn dataset can be found [here](https://github.com/ctlearn-project/ctlearn/blob/6b41352df6ef2a197fe31d9a656147f620a62ac4/config/v_0_3_0_benchmarks/README.md). Bilinear interpolation and rebinning seems to perform the best, while bicubic and nearest seems to perform the worst. The performance of oversampling is in-between.

Shifting methods:
We implemented two shifting methods, which are used for several hexagonal convolution packages. The first one is **axial addressing**, which is used in [IndexedConv](https://github.com/IndexedConv/IndexedConv).
This is the axial addressing of the LSTCam with dummy data (the index of the camera pixel is set as charge intensity): 
LSTCam_axial_addressing

@vuillaut @mikael10j  Is this orientation of the axial addressing of the LSTCam the same, which is used in IndexedConv? Did you used [ctapipe/instrument/camera.py](https://github.com/cta-observatory/ctapipe/blob/master/ctapipe/instrument/camera.py) to perform the axial addressing and get the index matrix? Is there anything more needed for apply the IndexedConv packages?

The second method is called **image_shifting** and basically shift every second column of about a half of a hexagon length. This procedure is used by [HexagDly](https://github.com/ai4iacts/hexagdly) and maybe [HexaConv](https://github.com/ehoogeboom/hexaconv)(?).


dl1_data_handler/image_mapper.py
```diff
@@ -1,19 +1,915 @@
...
+
+            map_method = self.mapping_method[camtype]
+            if map_method not in ['oversampling', 'rebinning', 'nearest_interpolation', 'bilinear_interpolation',
+                                  'bicubic_interpolation', 'image_shifting', 'axial_addressing']:
+                raise ValueError("Hex conversion algorithm {} is not implemented.".format(map_method))
+            elif map_method in ['image_shifting', 'axial_addressing'] and camtype in ['ASTRICam', 'CHEC', 'SCTCam']:
+                raise ValueError(
+                    "{} (hexagonal convolution) is not available for square pixel cameras.".format(map_method))
+
+            if map_method not in ['oversampling', 'image_shifting', 'axial_addressing']:
```
aribrill:
```suggestion
            if map_method in ['rebinning', 'nearest_interpolation', 'bilinear_interpolation', 'bicubic_interpolation']:
```

dl1_data_handler/image_mapper.py
```diff
@@ -1,19 +1,915 @@
...
+                    for k in np.arange(0, output_dim, 1):
+                        for l in np.arange(0, weights.shape[3], 1):
+                            if weights.shape[3] == 3:
+                                mapping_matrix3d[corner_indexes[i][k][j][l]][k + pad][j + pad] = weights[i][k][j][
+                                                                                                         l] / 4
+                            elif weights.shape[3] == 4:
+                                mapping_matrix3d[corner_indexes[k][j][i][l]][k + pad][j + pad] = weights[k][j][i][
+                                                                                                         l] / 4
+
+        # Cutting the mapping table after num_pixels+1, since the virtual pixels have intensity zero.
```
aribrill:
```suggestion
        # Cutting the mapping table after num_pixels, since the virtual pixels have intensity zero.
```

dl1_data_handler/reader.py
```diff
@@ -312,14 +315,14 @@ def _get_image(self, filename, tel_type, image_index):
         length = [x['num_pixels'] for x
                   in f.root.Telescope_Type_Information.where(query)][0]
         num_channels = len(self.image_channels)
-        vector = np.empty(shape=(length + 1, num_channels), dtype=np.float32)
+        vector = np.empty(shape=(length, num_channels), dtype=np.float32)
         # An "empty" pixel at index 0 is used to fill blank areas in image
         vector[0, :] = 0.0
```
aribrill: Delete lines 319-320

dl1_data_handler/image_mapper.py
```diff
@@ -1,19 +1,915 @@
...
+class ImageMapper:
+
+    def __init__(self,
+                 camera_types=None,
+                 mapping_method=None,
+                 padding=None,
+                 interpolation_image_shape=None,
+                 rotate_camera=False,
+                 mask_interpolation=False,
+                 channels=1):
```
aribrill:
```suggestion
                 channels=['charge']):
```

dl1_data_handler/image_mapper.py
```diff
@@ -1,19 +1,915 @@
...
+                self.image_shapes[camtype] = (
+                    self.image_shapes[camtype][0] + self.default_pad * 4,
+                    self.image_shapes[camtype][1] + self.default_pad * 4,
+                    self.image_shapes[camtype][2]
+                )
+
+            self.image_shapes[camtype] = (
+                self.image_shapes[camtype][0],
+                self.image_shapes[camtype][1],
+                channels  # number of channels
```
aribrill:
```suggestion
                len(channels)  # number of channels
```

dl1_data_handler/reader.py
```diff
@@ -312,14 +315,14 @@ def _get_image(self, filename, tel_type, image_index):
...
         # If the telescope didn't trigger, the image index is 0 and a blank
         # image of all zeros with be loaded
         for i, channel in enumerate(self.image_channels):
-            vector[1:, i] = record[channel]
-        image = self.image_mapper.map_image(vector, tel_type)
+            vector[0:, i] = record[channel]
```
aribrill:
```suggestion
            vector[:, i] = record[channel]
```

dl1_data_handler/image_mapper.py
```diff
@@ -1,19 +1,915 @@
...
+    def get_virtual_pixels(x_ticks, y_ticks, x, y):
+        gridpoints = np.array(np.meshgrid(x_ticks, y_ticks)).T.reshape(-1, 2)
+        gridpoints = [tuple(l) for l in gridpoints.tolist()]
+        virtual_pixels = set(gridpoints) - set(zip(x, y))
+        virtual_pixels = np.array(list(virtual_pixels))
+        return virtual_pixels
+
+    @staticmethod
+    def normalize_mapping_matrix(mapping_matrix3d, num_pixels):
+        norm_factor = np.sum(mapping_matrix3d[1:]) / float(num_pixels)
```
aribrill:
```suggestion
        norm_factor = np.sum(mapping_matrix3d) / float(num_pixels)
```

dl1_data_handler/image_mapper.py
```diff
@@ -1,19 +1,915 @@
...
+        mapping_matrix3d[1:] /= norm_factor
```
aribrill:
```suggestion
        mapping_matrix3d /= norm_factor
```

aribrill:
As a note, there are a couple of other outstanding issues relating to image mapping. I think these could be handled in separate PRs.

Add method to return unmapped vectors: #49 
Fix bug when both normalizing and masking: https://github.com/ctlearn-project/ctlearn/issues/85

Masking means to explicitly zero out all output pixels not nearest to positions of hexagonal camera pixels. The purpose is to remove an edge artifact caused by interpolation between real pixels and 0-valued artificial pixels just outside the camera, which would primarily affect partial/off-axis events. The problem is that doing this breaks the strict normalization of the mapping table. So far in practice the effect of this artifact seems minimal, as does the effect of masking in general.

aribrill:
I added a Jupyter notebook demonstrating the performance of the image mapper.

GitHub's notebook viewer isn't working for me. If you have the same problem, nbviewer should work:
[dl1_data_handler_demo](https://nbviewer.jupyter.org/github/cta-observatory/dl1-data-handler/blob/7e9c336307ccdc9ed6f44df38a090c823d0dbcab/notebooks/dl1_data_handler_demo.ipynb)
[test_image_mapper](https://nbviewer.jupyter.org/github/cta-observatory/dl1-data-handler/blob/5f1279b7b8e4f7271bfca404d0e4b73999332e2b/notebooks/test_image_mapper.ipynb)

dl1_data_handler/image_mapper.py
```diff
@@ -1,19 +1,915 @@
...
-        # Fill an array of the right size with copies of the vector
-        return np.resize(vector, self.image_shape[tel_type])
+
+        for camtype in self.camera_types:
+            # Get a corresponding CameraGeometry
+            self.camera_geometries[camtype] = CameraGeometry.from_name(camtype)
+
+            # Rotate it if necessary
+            #if camtype in ['LSTCam', 'NectarCam', 'MAGICCam']:
+                #self.camera_geometries[camtype].rotate(90.0 * u.deg - self.camera_geometries[camtype].pix_rotation)
```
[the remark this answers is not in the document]
TjarkMiener: Actually these lines can be deleted! The coordinates of the ctapipe-extras are align and straight to use for our purpose! Ctapipe-extras are shipping the coordinates together with the rotation angle. To get the original orientation, you have to apply the rotation to the coordinates.

dl1_data_handler/image_mapper.py
TjarkMiener: Yes!

dl1_data_handler/reader.py
TjarkMiener: Ok!

dl1_data_handler/reader.py
TjarkMiener: yes :D

dl1_data_handler/image_mapper.py
TjarkMiener: Good catch!

dl1_data_handler/image_mapper.py
TjarkMiener: Good catch again!

TjarkMiener:
@aribrill The mapping tables are sparse 3D matrices with the shape (number of camera pixels, output/image size, output/image size). The mapping tables contain weights, which are needed to compute the intensity of the output pixels. The intensity is a linear combination of the intensities of the camera pixels. The shape of the mapping tables is 100% needed for the conversion methods, but not for the shifting methods. Here the camera pixels will only be shifted, which means we don't have a linear combination and the weights, we are storing, are either 0 or 1. I implemented the shifting method the same way as the conversion method to have a clean map_image() function. However, I now strongly believe we can shrink the mapping table into one single matrix for the shifting method. Rather than storing weights, we will store the indexes. Is this is the input, which is needed for the IndexedConv packages, @vuillaut @mikael10j?

mikael10j:
> Rather than storing weights, we will store the indexes. Is this is the input, which is needed for the IndexedConv packages, @vuillaut @mikael10j?

In a general way, you need to provide the IndexedConv with the list of the neighbours (their indices) of each pixel of interest. To compute this list for hexagonal pixel images, the package has already functions based on the axial addressing system. If the indices you're storing respect the axial addressing system, you can use them directly, otherwise you need to create the functions to extract the list of neighbours. 
For the record, we have added to the package the functions to build this list based on Delaunay tessellation. We should release it in the following weeks.

mikael10j:
> @vuillaut @mikael10j Is this orientation of the axial addressing of the LSTCam the same, which is used in IndexedConv? Did you used [ctapipe/instrument/camera.py](https://github.com/cta-observatory/ctapipe/blob/master/ctapipe/instrument/camera.py) to perform the axial addressing and get the index matrix? Is there anything more needed for apply the IndexedConv packages?

The orientation looks like the one we have. To perform the axial addressing we've so far used functions created by Pierre and I wasn't aware of the ability of ctapipe to do it. Thank you for pointing it out. I need to check if the result is the same.

mikael10j:
> The orientation looks like the one we have.

Actually our orientation is not the same.
Here is the default geometry in ctapipe:

> CameraGeometry: "LSTCam"
>  - num-pixels: 1855
>  - pixel-type: hexagonal
>  - sensitive-area: 4.016016000618191 m2
>  - pix-rotation: 100.893 deg
>  - cam-rotation: 0.0 deg
image

In our case we derotate the image first:
>CameraGeometry: "LSTCam"
 >  - num-pixels: 1855
 >  - pixel-type: hexagonal
 >  - sensitive-area: 4.016016386167223 m2
 >  - pix-rotation: 0.0 deg
 >  - cam-rotation: -100.893 deg
image

and then use a function created by Pierre (in hipecta) to find the position of the pixels in the axial addressing system. From this we create the index matrix of the image:
image
fake pixels have an index of -1.

MaxNoe:
As discussed here: https://github.com/cta-observatory/ctapipe/issues/255

The derotation should probably not be done. If the LST camera has a flat top, the coordinate system should be in the HESS MC coordinate system.

TjarkMiener:
Thanks @mikael10j for clarifying the orientation! The commit bc2db97 solve the orientation discrepancy. I added a function to retrieve the index matrix (input of IndexedConv) in commit bd1e66c.

The rotation/de-rotation can be quite confusing! Thanks for linking us to the useful conversation, @MaxNoe! To apply the axial addressing, we have to align the hexagonal grid with the x- and y-axes. This is why it's obligatory to de-rotate the affected camera types. The de-rotation is also obligatory for all conversion methods, but here we added a second rotation, which rotate the camera back to the original (right) orientation.

mikael10j:
Hi @TjarkMiener , I've compared the index matrix returned by ImageMapper and the one that we build and it appears that the former is flipped compared to the latter.
`im = ImageMapper(camera_types=['LSTCam'], mapping_method={'LSTCam': 'axial_addressing'})
dl1_index_matrix = im.get_indexmatrix('LSTCam')
plt.imshow(dl1_index_matrix)`
image
`plt.imshow(ic_index_matrix.squeeze(0).squeeze(0))`
image
`torch.allclose(torch.tensor(np.ascontiguousarray(np.flip(dl1_index_matrix, 0)), dtype=torch.float32), ic_index_matrix.squeeze(0).squeeze(0))`
True

TjarkMiener:
Thanks, @mikael10j! In CTLearn the indexes of our images used to start from bottom left corner, while the indexes of your images start from top left corner. All images produced by the DL1DH should have that flip then! What would be the smartest way of solving this? Adding an option to the ImageMapper that will perform the flip on the output images? We can also perform the flip on the mapping tables (in the init), but there we have to perform the flip "number of camera pixels" times. @aribrill @bryankim96 @nietootein 

I'm getting the same orientation (in CTLearn indexing):
LSTCam_axial

TjarkMiener:
We adopted your convention, @mikael10j @vuillaut. Now the output of map_image() and get_indexmatrix() should be exactly what you had before. Can you please check?

mikael10j:
Thank you @TjarkMiener , I check today.

mikael10j:
Looks good !

PR #67, opened 2019-06-18
Update data transform for gammalearn

mikael10j:
- Fix class_label dtype for gammalearner
- Transpose image


dl1_data_handler/transforms.py
```diff
@@ -153,12 +153,21 @@ def __call__(self, example):
         array = []
         for i, (val, des) in enumerate(zip(example, self.description)):
             if des['base_name'] == 'image':
-                image = val
+                image = val.T
```
vuillaut: why?

mikael10j: because the reader returns the images with shape (num_pixels, channels)

PR #75, opened 2019-10-03
Added a new mapping method 

TjarkMiener:
I "added" a new mapping method called 'IndexedConv', which will be the input (unmapped vector) for the [IndexedConv](https://github.com/IndexedConv/IndexedConv) package. With get_indexmatrix() one can retrieve the index matrix (as before), which is also needed for IndexedConv. The output for 'axial_addressing' in the ImageMapper is now a **2D** images, which can be used for masked convolution layers.


TjarkMiener:
Done! Thanks for the review.

Issue #74, opened 2019-10-02
Update to ctapipe v0.7.0 


TjarkMiener:
I tried to update to `ctapipe v0.7.0` (or/and current master of ctapipe). So far, I'm able to produce the h5 files without any breaks. However, my configurations of the `ctapipe CameraCalibrater` or of the `ctapipe GainSelector` are not right. Can anyone (more familiar with ctapipe) please have a look into this? You can find the code [here](https://github.com/cta-observatory/dl1-data-handler/tree/update_ctapipe). Thanks in advance! The 'charge' channel seems to be working fine, while the 'pulse_time' channel is messed up.
![LSTCam_gamma_66_charge](https://user-images.githubusercontent.com/37835610/66048971-d5363f00-e52a-11e9-9208-6ab7fb8341f2.png)
![LSTCam_gamma_66_time](https://user-images.githubusercontent.com/37835610/66048972-d5363f00-e52a-11e9-8252-fdd25b9f2e62.png)

vuillaut:
Hi @TjarkMiener.
Thank you for looking into this. I was thinking about it but really didn't have time.
Not sure the pulse time map is messed-up without the scale. The way the time map is computed has changed and I think the color representation is misleading you here.

TjarkMiener:
Thanks for the quick response, @vuillaut! It seems you are right. It's a matter of color representation. However, we are puzzled, why we are getting the time arrival values in this ranges?

I plotted the same event with a colorbar:
![LSTCam_gamma_66_time](https://user-images.githubusercontent.com/37835610/66057691-71ffd900-e539-11e9-8932-8e307c15c016.png)

I plotted a histogram of the time arrival of the same event:
![LSTCam_histogram_gamma_66_time](https://user-images.githubusercontent.com/37835610/66057576-411fa400-e539-11e9-86bc-205d9cd08429.png)

I plotted the same event (in logscale) with a colorbar:
(Here one can spot the shower)
![LSTCam_logscale_gamma_66_time](https://user-images.githubusercontent.com/37835610/66057726-8512a900-e539-11e9-8df7-ec0d925dcb0d.png)

kosack:
If you think the values are not correct, can you post an issue on the ctapipe issue tracker? Not so many devs read this one.  

By the way, you can just set `display.norm = 'log'` to show a log-scale image, rather than modifying the values.

vuillaut:
[...]

PR #76, opened 2019-11-21
Update to ctapipe current master

TjarkMiener:
Created to address #74


TjarkMiener:
I updated the README with an installation instruction, which is working. I will open a new issue about the bug in the installation.

Issue #77, opened 2019-12-02
Wrong mapping for FlashCam


TjarkMiener:
There is a mapping bug for FlashCam, while using the current ctapipe and dl1-data-handler master. 
Example gamma event: 
![FlashCam_oversampling](https://user-images.githubusercontent.com/37835610/69952167-e4b81100-14f6-11ea-9eed-c543b95fc2c8.png)
This is due to the discrepancy of the pixel positions (in the following the pixel intensity is pixel index) from the ctapipe (`ctapipe.instrument.camera.CameraGeometry`)
![ctapipe_FlashCam](https://user-images.githubusercontent.com/37835610/69952384-56905a80-14f7-11ea-8a23-1ed5edb9908b.png)
and from the dl1 h5 files (`subarray.tel[tel_id].camera.pix_x.value`)
![h5_pixposFlashCam](https://user-images.githubusercontent.com/37835610/69952620-dd453780-14f7-11ea-9238-d577acb63489.png)
The ImageMapper class is currently reading the pixel positions from ctapipe, which leads to the scattered events above. 

I would suggest to read the pixel positions from the h5 files in the reader and pass the pixel positions to the ImageMapper. If there is no pixel positions passed to the ImageMapper, the ImageMapper should get the pixel positions from ctapipe. This also make it possible to call the dl1 reader and ImageMapper without ctapipe dependency.

PR #78, opened 2019-12-02
Fixed mapping bug for FlashCam

TjarkMiener:
Created to address #77


dl1_data_handler/image_mapper.py
```diff
@@ -72,16 +71,36 @@ def __init__(self,
...
+                    rotation_matrix = np.matrix([[np.cos(rotation_angle), -np.sin(rotation_angle)],
+                                                 [np.sin(rotation_angle), np.cos(rotation_angle)]], dtype=float)
+                    self.pixel_positions[camtype] = np.squeeze(np.asarray(np.dot(rotation_matrix, self.pixel_positions[camtype])))
+                    if self.rotate_camera:
+                        self.rotate_back_angle = 90.0-camgeo.pix_rotation.deg
+            else:
+                self.pixel_positions[camtype] = pixel_positions[camtype]
+                self.num_pixels[camtype] = pixel_positions[camtype].shape[1]
+                if camtype in ['LSTCam', 'NectarCam', 'MAGICCam'] and self.rotate_camera:
+                    self.rotate_back_angle = rotate_back
```
aribrill: Should rotate_back_angle be a dictionary? If there are multiple tel types, only some of which should be rotated, this will set the angle to the value for the last tel type, whatever that happens to be.

dl1_data_handler/image_mapper.py
```diff
@@ -513,8 +526,8 @@ def generate_table(self, camera_type):
         if self.mask and map_method in ['bilinear_interpolation', 'bicubic_interpolation']:
             mapping_matrix3d = self.apply_mask_interpolation(mapping_matrix3d, nn_index, num_pixels, pad)
         # Rotating the camera back to the original orientation
-        if self.rot_cam and camera_type in ['LSTCam', 'NectarCam', 'MAGICCam'] and map_method not in ['image_shifting', 'axial_addressing', 'indexed_conv']:
-             mapping_matrix3d = self.rotate_mapping_table(mapping_matrix3d,90.0-CameraGeometry.from_name(camera_type).pix_rotation.deg)
+        if self.rotate_camera and camera_type in ['LSTCam', 'NectarCam', 'MAGICCam'] and map_method not in ['image_shifting', 'axial_addressing', 'indexed_conv']:
+             mapping_matrix3d = self.rotate_mapping_table(mapping_matrix3d,self.rotate_back_angle)
```
aribrill:
```suggestion
             mapping_matrix3d = self.rotate_mapping_table(mapping_matrix3d, self.rotate_back_angle)
```

dl1_data_handler/image_mapper.py
TjarkMiener: Right! rotate_back_angle should be a dict.

TjarkMiener:
This PR would be ready to merge.  We thought on having two installations for the DL1DH, because the DL1 writing will be done in ctapipe in the future. One light installation without ctapipe dependence, which will be the default one and one full installation, which gives the user full access to DL1DH. The DL1DataReader and it helper classes will be usable with the light installation without ctapipe.
Please let me know what you think, @aribrill @vuillaut @nietootein  @mikael10j .

PR #85, opened 2020-03-11
dl1-data-handler v0.8.2

TjarkMiener:


dl1_data_handler/utils.py
```diff
@@ -85,3 +85,46 @@ def clean(img):
...
+    def leak(geom, img, lkval, lknum, **opts):
+        cleanmask = cleaning.tailcuts_clean(geom, img, **opts)
+        mask = False
+        if any(cleanmask):
+            # ctapipe v0.7.0
+            print(leakage(geom, img, cleanmask)['leakage{}_intensity'.format(lknum)])
+            mask = leakage(geom, img, cleanmask)['leakage{}_intensity'.format(lknum)] <= lkval
+        return mask
+            
+    leakage_mask = np.array([leak(geom, image, leakage_value, leakage_number, **opts) for image in images])
```
mikael10j: Hi @TjarkMiener , to be more efficient, you can use numpy.apply_along_axis instead of a for loop:
```Python
    def leak(img):
        cleanmask = cleaning.tailcuts_clean(geom, img, **opts)
        mask = False
        if any(cleanmask):
            # ctapipe v0.7.0
            print(leakage(geom, img, cleanmask)['leakage{}_intensity'.format(leakage_number)])
            mask = leakage(geom, img, cleanmask)['leakage{}_intensity'.format(leakage_number)] <= leakage_value
        return mask
    leakage_mask = np.apply_along_axis(leak, 1, image)
```

TjarkMiener: Thanks @mikael10j! I will adopt your changes. I will also add a filter for the image intensity after cleaning. What do you think of renaming `utils.py` to `filters.py`, since it only contains filters?

TjarkMiener:
Hi @mikael10j @vuillaut @aribrill @nietootein,
The last commit added a transformation from alt/az to delta alt/az, which is the difference of alt az of the arrival direction and the alt/az of the telescope pointing (both in rad). The telescope pointing is read from the h5 file and stored in a class variable of the reader, so you can easily transform the prediction "back" to original alt/az. What do you think of this transformation?

PS: The offset we saw for pointlike gammas was due to 0.4 deg pointing offset, which I didn't include (I inserted the wrong telescope pointing by hand). Now with this transformation, the pointing is read automatically and everything is working as it should be (no offset is observed).

TjarkMiener:
This PR is ready from my side. Any further comments?

PR #87, opened 2020-05-15
Passing reader.py's camera list to image_mapper.py instead of the default init

sahilyadav27:


dl1_data_handler/reader.py
```diff
@@ -198,8 +198,7 @@ def __init__(self,
                     self.pixel_positions[cam] = np.squeeze(np.asarray(np.dot(rotation_matrix, self.pixel_positions[cam])))
         if rotate_back:
             mapping_settings['rotate_back'] = rotate_back_angle[cam]
-        self.image_mapper = ImageMapper(pixel_positions=self.pixel_positions,
-                                        **mapping_settings)
+        self.image_mapper = ImageMapper(pixel_positions=self.pixel_positions, camera_types = cameras, **mapping_settings)
```
TjarkMiener, line 201: Can you add the returns here? It would be easier to read.

TjarkMiener: And you need to init 'cameras' with None before line 183. Otherwise it will break if  `/Telescope_Type_Information` is not in h5.

sahilyadav27: Sure!

Issue #92, opened 2020-11-17
Switch to use ctapipe-stage1?


maxnoe:
Hi,

As the ctapipe stage1 tool is now available and writes images and parameters in a standard way, 
would you consider using that instead of duplicating efforts?

Is there something missing from the stage1 tool you need?

Do you need some support that would enable you to make the switch?

vuillaut:
Hi

Yes it has always been an explicit goal to switch to stage1 tool at some point, keeping here only  the reader for DL studies.

This requires to:
- test thoroughly the stage1 DL1 (anyone ?)
- modify the reader, which will require some work

maxnoe:
Yes, also note that ctapipe 0.10 is imminent (today or tomorrow) with fixes also for the stage1 output. So it would be great if you could test it as soon as 0.10 is out.

maxnoe:
For the reading part, ctapipe 0.10 also includes the `DL1EventSource`, which can be used to iterate event wise.

But for most applications, I think you don't actually want that and instead just use `ctapipe.io.astropy_helpers.h5_table_to_astropy` to read a whole table into an astropy table, e.g. all images of one telescope.

kosack:
Note that that function is aliased to `ctapipe.io.read_table`. Soon I'll add a tutorial on working with the DL1 data files, including merging datasets, etc, which could be useful for you

nietootein:
Thanks for opening this issue, @maxnoe !

As @vuillaut said, we'd like to switch to stage1 at some point. 

While waiting for that tool, @TjarkMiener and @sahilyadav27 have been expanding the writer so we can now dump MAGIC data in ROOT format to hdf5 (see #90), aiming at implementing similar functionality for VERITAS and H.E.S.S. data as well (with the purpose of using our DL frameworks on data from current gen IACTs). Is this feature something envisioned for stage1?

Additionally, @LucaRomanato and @TjarkMiener have recently implemented the calculation and addition of image parameters to the writer (see #91), not only for CTA cameras.

I wonder if some of these and other functionalities (e.g. imagemapper transformations) could be of interest for stage1 users other than us, so to transfer them there?

maxnoe:
> Is this feature something envisioned for stage1?

Yes, you just need to implement an event source in a package called `ctapipe_io_<name>` and the stage1 tool will automatically choose a compatible event source if available.

[...]

PR #94, opened 2020-12-08
Include parameters in the data format 

TjarkMiener:
Following up on @LucaRomanato's PR #91.

Also some additional fields have been added to the format, which makes some transformations in transform.py redundant. This will give an overall speedup for the training process in CTLearn.

The reader will be updated in a separate PR.


kosack:
What is the difference between this and ctapipe_io_magic (https://github.com/cta-observatory/ctapipe_io_magic)? If you are developing a ctapipe EventSource, it should be a separate package like that one, and then it becomes an automatically detected plugin.  After that, just using code like this will work generally for all formats supported (Sim-R0, LST-R1, DL1, MAGIC, etc.)

```py3
with EventSource(filename) as source:
    for event in source:
         process(event)
```
Will work no matter what format the file is in, as long as there is an  io plugin for it installed.  So if you separate this code which I would rename "MAGICDL1EventSource", if it generates data at the DL1 level), and either replace the old MagicEventSource (if it is not longer maintained), or make a new one, then you will save a lot of effort and it will be re-usable by everyone.

TjarkMiener:
@kosack yes that's a very good suggestion! I think making it a separate ctapipe_io_MAGICDL1EventSource is a little bit overkill, since it only feature Deep Learning relevant information of the DL1 level. I will contacted the maintainers of ctapipe_io_magic and suggest to include it as a subclass of the main more general one. Let's see what they are thinking about this.

TjarkMiener:
I contacted the MAGIC DL convenors about moving this EventSource as a subclass to the ctapipe_io_magic package. In the meantime, can this PR get merged? Anything missing for v0.9.0, @nietootein @rlopezcoto?

dl1_data_handler/writer.py
```diff
@@ -465,52 +645,139 @@ def _create_event_table(self, subarray):
...
+                                                         "Parameter table of {} parameters, algorithm: {}, args: {}".format(tel_name,
+                                                                  self.cleaning_settings[index_parameters_table -1]['algorithm'],
+                                                                  self.cleaning_settings[index_parameters_table -1]['args']),
+                                                        filters=self.filters,
+                                                        expectedrows=expected_rows)
+
+                # Place blank image at index 0 of all image tables
+                parameter_row = parameter_table.row
+
+                parameter_row['event_index'] = -1
```
rlopezcoto, line 742: better to initialize values to `np.nan` or `None`, see the initialization of all these quantities in:
https://github.com/cta-observatory/cta-lstchain/blob/master/lstchain/io/lstcontainers.py

config/example_config.yml
```diff
@@ -1,17 +1,40 @@
...
+    image_extractor_settings:
+        algorithm: 'LocalPeakWindowSum'
+        args:
+            window_shift: 4
+            window_width: 8
+    # parameters0 table, cleaning_algo and cleaning_args
+    cleaning_settings:
+        algorithm: 'tailcuts_clean'
+        args:
+            picture_thresh: 7
```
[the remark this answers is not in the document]
TjarkMiener: Solved

dl1_data_handler/writer.py
TjarkMiener: I agree, it's better to use `np.nan` for initialization, but those are stored in the file, i.e. the data types have to match the with the real quantities. It's needed to construct the tables before.

TjarkMiener: Solved. Thanks for the hint @rlopezcoto :)

TjarkMiener:
Hi @kosack! Following up on the discussion above, I found out that the ctapipe_io_magic is actually maintained in [this gitlab repo](https://gitlab.mpcdf.mpg.de/ievo/ctapipe_io_magic). I can ping the maintainers asking to synchronise the gitlab with the github repo of cta-observatory, so it could be directly detected by ctapipe as an external EventSource. So, the plan would be to leave the MAGIC related code here for the records and move to this EventSource once the transition of the ctapipe_io_magic to github is done and we adopt the stage1 for DL #92.

dl1_data_handler/writer.py
maxnoe: That is an integer, -1 is the correct choice as there is no nan for integers and putting nan will result in the value being stored in the wrong data type.

rlopezcoto (approved):

nietootein (approved):

Issue #96, opened 2021-01-15
read the parameter tables in the DL1DH reader


TjarkMiener:
The reader need some adjustment to incorporate with #94. Parts are also needed for #92.

TjarkMiener:
Solved in v0.10.0

PR #97, opened 2021-01-15
Upgrade reader.py in order to load image parameters

LucaRomanato:
In this PR I implemented the adjustment to incorporate PR #94 asked in issue #96.
Now the reader.py load correctly images for mode mono and stereo. In addiction, in mono mode, reader correctly load image parameters for each image.
Also filters are updated.

### DIFF
--- a/dl1_data_handler/reader.py
+++ b/dl1_data_handler/reader.py
@@ -136,6 +136,14 @@ class DL1DataReader:
             selected_nrows &= self._select_event(f, event_selection)
             selected_nrows = list(selected_nrows)
 
+            # got parameters table selected
+            if image_selection_from_file != {}:
+                algorithm = image_selection_from_file[next(iter(image_selection_from_file))]['algorithm']
+                parameters_table = f.root['/Parameters' + str(algorithm)][self.tel_type]
+            else:
+                algorithm = 0
+                parameters_table = f.root['/Parameters0'][self.tel_type]
+
             # Make list of identifiers of all examples passing event selection
             if self.mode in ['stereo', 'multi-stereo']:
                 example_identifiers = [(filename, nrow) for nrow
@@ -143,12 +151,6 @@ class DL1DataReader:
             elif self.mode == 'mono':
                 example_identifiers = []
                 # Determinate which parameters table loads due to user settings in the config file
-                if image_selection_from_file != {}:
-                    algorithm = image_selection_from_file[next(iter(image_selection_from_file))]['algorithm']
-                    parameters_table = f.root['/Parameters' + str(algorithm)][self.tel_type]
-                else:
-                    algorithm = 0
-                    parameters_table = f.root['/Parameters0'][self.tel_type]
                 field = '{}_indices'.format(self.tel_type)
                 selected_indices = f.root.Events.read_coordinates(selected_nrows, field=field)
                 for tel_id in selected_telescopes[self.tel_type]:
@@ -512,12 +514,13 @@ class DL1DataReader:
                                                             tel_type)
                 example.extend(tel_type_example)
 
-        # Load parameters
-        with lock:
-            parameters = self.files[filename].root['/Parameters' + str(self.algorithm)][self.tel_type]
-            for column in self.training_parameters:
-                dtype = parameters.cols._f_col(column).dtype
-                example.append(np.array(parameters[image_index][column], dtype=dtype))
+        # Load parameters, working only with mono mode at the moment
+        if self.mode == "mono":
+            with lock:
+                parameters = self.files[filename].root['/Parameters' + str(self.algorithm)][self.tel_type]
+                for column in self.training_parameters:
+                    dtype = parameters.cols._f_col(column).dtype
+                    example.append(np.array(parameters[image_index][column], dtype=dtype))
 
         # Load event info
         with lock:
--- a/dl1_data_handler/reader.py
+++ b/dl1_data_handler/reader.py
@@ -21,6 +21,7 @@ class DL1DataReader:
                  mode='mono',
                  selected_telescope_type=None,
                  selected_telescope_ids=None,
+                 selected_parameters=None,
                  selection_string=None,
                  event_selection=None,
                  image_selection=None,
@@ -55,6 +56,9 @@ class DL1DataReader:
         self.telescopes = {}
         if selected_telescope_ids is None:
             selected_telescope_ids = {}
+
+        if selected_parameters is None:
+            selected_parameters = {}
 
         if event_selection is None:
             event_selection = {}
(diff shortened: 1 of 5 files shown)


TjarkMiener:
hm i pretty much surprised that I already comment on those changes. Even reading my comments, I can't remember :D 

Thank you so much for this PR, though! Great to see this nice progress and the hard work you put into this! This would be also very useful for the migration to the stage1 tool #92.

LucaRomanato:
Thank you @TjarkMiener !
maybe I just hacked your github profile and commented my changes ( just kidding ). I hope this changes will be usefull for you ^^

dl1_data_handler/reader.py
```diff
@@ -176,9 +176,10 @@ def __init__(self,
                         for parameter_name in self.training_parameters:
                             if parameter_name != 'event_index':
                                 if parameter_name == 'hillas_log_intensity':
-                                    param = [np.log10(x['hillas_intensity']) for x in parameters_table if x > -1]
-                                    param_list.append(-1)
-                                    param_list.extend(param)
+                                    param = [np.log10(x['hillas_intensity']) for x in parameters_table if
```
maxnoe, line 179: Do you really have a python for loop here over *every* event? And why do you need to insert a -1 at the front?

LucaRomanato: Hello @maxnoe, thanks for you comment.
First of all this _if_ ( i.e. if parameter_name == 'hillas_log_intensity' ) will be removed when a new production will be processed with the dl1dh writer.  I needed to do this because during the last production a bug had not discovered  yet and the parameters values were calculated with a wrong log base. Since the IT cluster failed and my I don't have time ( I'm a master student and my thesis period is about to end ), I cannot re-do the production so I decided to fix it coding something. 
Then, I put a -1 as first list element because tables were inizialized with -1 ( see #94 ). If i don't put it the other parameter lists will have a different length and this will cause errors in the ctlearn models. 
In order to get each parameter list in its entirety I did in this way. Parameters are more or less 30 ( but only 12 are usually used ) and computational time was not affected ( the time difference with and without parameters is more or less 2 minutes ) so I left this coding solution but please, if you have any suggestions or adivices I will be very happy to know and learn something new

maxnoe: Some points:

* Putting the -1 as first element will only ever work if and only if only the first event has ` -1`.  That is certainly not always the case.
* Almost never loop over numpy arrays. Numpy functions work on arrays and are much faster then applying them in a loop on each value. So if you want to transform with log, just call `np.log10(x['hillas_intensity'])`. If you want to replace invalid values with -1, use masking:
```
valid = parameters_table['hillas_intensity'] > 0
log10_intensity = np.empty(len(parameters_table))
log10_intensity[valid] = np.log10(parameters_table['hillas_intensity'])
log10_intensity[~valid] = -1 
```

TjarkMiener: I need to go through this in more detail, but it seems that this belongs in the transforms.py as a transformation. Something similar was done with the energy in the past. One question for grouping the parameters in a list. This is needed to directly concat the parameters list with the CNN output? Would it be possible to also open a PR in CTLearn with your changes or if it isn't ready maybe you can point me to changes in your fork? Thanks!

LucaRomanato: let me answer to @TjarkMiener before: This if-else with np.log10() **is temporary** because from the next dl1dh production the `np.log10()` will be directly performed inside the `writer.py`. If you re-do the production now ( if possible ) you can simply delete it without put it in the transforms.py script.
Regarding the parameters. As I have implemented the CTLearn model, concatenating the parameter lists is strictly necessary otherwise there would be errors. You can see the model here ( it works but any advice is pretty welcomed ): https://github.com/LucaRomanato/ctlearn/blob/multitask/ctlearn/default_models/cnn_fcnn.py

LucaRomanato: Thanks for your suggestion @maxnoe !

regarding the first point, I'm sure the first value is -1 because this is the value used for initialize tables in the dl1dh `writer.py`. With this for loop my intention is to load all the selected parameter ( such as hillas_intensity ) values in the table except the first (  `if x['hillas_intensity'] > -1`  is because hillas_intensity is always positive, only the first value is negative in every case), and then perform the `np.log10()`. 
example of a parameter list in the h5 file: [-1, 2 ,3 ,4 ,5] where -1 is the parameter value for event_index = -1, 2 is for event_index = 0 and so on.. 

For the second point I don't know if `parameters_table['hillas_intensity'] > 0` works because ` parameters_table = file.root.Parameters0.LST_LST_LSTCam`, I need to try it :) but separate the `np.log10() `and perform it after is a great advice for me, thank you!

maxnoe: > regarding the first point, I'm sure the first value is -1 because this is the value used for initialize tables in the dl1dh writer.py

You writer always writes an artificial first event with -1 as parameter? That is wrong.

TjarkMiener: There is an artificial blank images stored in the hdf5 referred as event_index -1. Telescopes that not trigger in stereo get this -1 as event_index to map it to this blank image. This is needed for the CNN-RNN model (performing stereo reconstruction), where you have to fill blank images depending on the event. For mono analysis this blank image is safely ignored.

dl1_data_handler/reader.py
```diff
@@ -140,13 +159,36 @@ def __init__(self,
...
+                                    param_list.append(param)
+                                else:
+                                    param = [x[parameter_name] for x in parameters_table]
+                                    param_list.append(param)
+                        for image_index, nrow in zip(img_ids[mask], np.array(selected_nrows)[mask]):
+                            temp_list = []
+                            for list_index in range(len(param_list)):
+                                params_value = param_list[list_index][image_index]
+                                temp_list.append(params_value)
+                            example_identifiers.append((filename, nrow, image_index, tel_id, temp_list))
```
TjarkMiener, line 191: The example_identifiers should only hold the identifiers and not any data. Therefore the parameters should merged in a list later i.e. def __getitem__(self, idx). Similar to the images. Also the transformation to log should go there.

dl1_data_handler/reader.py
```diff
@@ -295,8 +340,24 @@ def __init__(self,
...
                             'shape': (num_tels,) + col.shape[1:],
                             'dtype': col.dtype
-                            }
-                        )
+                        }
+                    )
+
+        # Add parameters info to description
+        # working only with mono at the moment
+        if self.mode == 'mono':
```
TjarkMiener, line 348: move this to the if/elif/else above

dl1_data_handler/reader.py
```diff
@@ -295,8 +340,24 @@ def __init__(self,
...
+            for col_name in self.training_parameters:
```
TjarkMiener, line 349: With this you are appending an unprocessed_example_description for each parameter selected. But what you want is that you have one unprocessed_example_description of parameters matching the shape of self.training_parameters. Similar to how different channels (charge+ peak_time) are handled in the images unprocessed_example_description.

TjarkMiener:
I git rebased this PR into PR #100 and resolved the open tasks.

PR #99, opened 2021-02-28
Fix counting of class examples for labels that are sequences

aribrill:
This commit provides for correct conversion to tuples when counting labels that include sequences such as (alt, az). This is required for full event reconstruction with CTLearn (ctlearn-project/ctlearn#132).

### DIFF
--- a/dl1_data_handler/reader.py
+++ b/dl1_data_handler/reader.py
@@ -482,7 +482,7 @@ class DL1DataReader:
         for idx in example_indices:
             example = self[idx]
             # Convert to a tuple to get a hashable key
-            group = self._totuple(example[idx] for idx in grouping_indices)
+            group = self._totuple(example[idx].tolist() for idx in grouping_indices)
             if group in group_nums:
                 group_nums[group] += 1
             else:
--- a/dl1_data_handler/reader.py
+++ b/dl1_data_handler/reader.py
@@ -456,6 +456,15 @@ class DL1DataReader:
 
         return example
 
+
+    # Convert a possibly-nested sequence to a tuple
+    def _totuple(self, a):
+        try:
+            return tuple(self._totuple(i) for i in a)
+        except TypeError:
+            return a
+
+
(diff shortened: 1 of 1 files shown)


dl1_data_handler/reader.py
```diff
@@ -456,6 +456,15 @@ def __getitem__(self, idx):
 
         return example
 
+
+    # Convert a possibly-nested sequence to a tuple
+    def _totuple(self, a):
+        try:
+            return tuple(self._totuple(i) for i in a)
+        except TypeError:
+            return a
```
TjarkMiener, line 465:
```suggestion
            return tuple(a)
```

TjarkMiener: That was needed for me (for breakdown by 'particletype'). Can you please double check?

To breakdown by 'direction', it was breaking in `log_examples()` from the ctlearn utils as mentioned in [CTLearn PR 132](https://github.com/ctlearn-project/ctlearn/pull/132).

TjarkMiener (changes requested):

dl1_data_handler/reader.py
aribrill: I fixed the problem instead by converting the numpy array to a scalar. That is cleaner since otherwise the scalar labels would be wrapped as a tuple.

TjarkMiener: Great! It's working fine! We probably want to have this fix on top of dl1dh v0.8.2, i.e. releasing dl1dh v0.8.3, to incorporate with ctlearn v0.5.0? Currently this PR want to merge into the master, which is dl1dh >v0.9.0.

TjarkMiener (approved):

dl1_data_handler/reader.py
aribrill: Great. I'll merge this PR into master now, and would you be able to create a v0.8.3 release using this branch?

PR #100, opened 2021-03-01
Migration to Stage1 format

TjarkMiener:
Hi all,

This PR adds the migration to the stage1 format. The code can be tested with the [stage1 notebook](https://github.com/cta-observatory/dl1-data-handler/blob/stage1/notebooks/dl1reading_stage1.ipynb). I git rebased PR #97 into this branch.

Changes w.r.t. the current (master) reader:

- Parent class for shared code. Each format has it own subclass with different example identifiers. However, the "output" of the reader (`__getitem__` return) is generic though. 
- Option to read/store the example identifiers in a pandas hdf5 file.
- Different pointing modes supported.
- Read images and/or image parameters.
- Read cleaned images.
- Construct the pyIRF simulated Table.
- Only have 'mono' and 'stereo' reading mode. The previous multi-stereo mode can be obtained with the stereo mode and selection of different telescope types.


dl1_data_handler/reader.py
```diff
@@ -306,6 +305,12 @@ def __init__(self,
                  transforms=None,
                  validate_processor=False
                 ):
+
+        # Import ctapipe DL1 reader
+        try:
+            from ctapipe.io import read_table # let us read full tables inside the DL1 output file
```
maxnoe, line 311: You require `ctapipe >= 0.10.5` in the setup.py. That provides read_table. Why is this try except needed?

TjarkMiener: Hi @maxnoe. Thanks for the catch! True, it isn't needed.

TjarkMiener:
Hi @nietootein. This PR is pending for a while and it was already used to produce the results for [2109.05809](https://arxiv.org/abs/2109.05809) and [2112.01828](https://arxiv.org/abs/2112.01828). For reproducibility purposes, can we merge this PR and release a new version with known issues (see #104)?

This PR requires CTLearn v0.5.1 [#136](https://github.com/ctlearn-project/ctlearn/pull/136)

nietootein:
Sorry for the latency, @TjarkMiener. PR merged.

PR #101, opened 2021-04-14
Add ability to write non-MC MAGIC data

astrojarred:
The main goal of this PR is to add the ability to read real (non-MC) MAGIC data for use with the DL1 data handler. Previously it was only possible to read in and convert MAGIC MC data.

The main roadblocks and their eventual solutions included in this PR are:

1. Determining if the input file is real data or data produced by MC simulations. This is now performed by checking the run header.
2. Sometimes the run number needed to be determined. Now the run number is run directly from the root file.
3. Many of the paths to the root data arrays in the` _generator` function were specific to MC data. To solve this we just created a property called `self.is_simulation` and put various if/else statements within the generator.
4. There was no headers container for real datas so we added a new class to `containers.py`. In addition to the class `MAGICMCHeaderContainer` to handle MC data, we added `MAGICHeaderContainer` for real data.

These changes can be tested by using some MAGIC data as input to the `write_data.py` script.


maxnoe:
Please consider working together with @IevgenVovk and @moritzhuetten to update https://github.com/cta-observatory/ctapipe_io_magic so it can be used instead.

It would be better for anyone if you would just provide a proper event source for MAGIC data for use with ctapipe.

astrojarred:
Hi @maxnoe, thanks for bringing this to my attention. I was unaware of the work being done by the others, it's looking great! This PR was submitted on behalf of the Deep Learning group of MAGIC with @TjarkMiener and @rlopezcoto to quickly produce some datasets for training. 

The three of us are now in touch with the people you mentioned, so we can hopefully find a way to contribute to ctapipe_io_magic.

TjarkMiener:
Hi @astrojarred @maxnoe. As discussed offline and in the LST-MAGIC call on Wednesday, Jarred and me would be contributing to the ctapipe_io_magic to update this package to the latest ctapipe version. Your help would be very appreciated, Max! Having a pipeline to dump MAGIC data in the stage1 format would be terrific!

For now, I would merge this PR and the 'magic_data' branch, so it can be used to produced first results.

TjarkMiener (approved):

maxnoe:
@TjarkMiener Sounds good. I am a bit swamped right now, but if you want to have a short meeting / coding sprint I would be open.

I can walk you through what was necessary for ctapipe_io_lst, but as I understood, the MAGIC situation is much easier, since you want to only start already at DL1a, right? Not with waveforms.

TjarkMiener:
Yes, correct! For ctapipe_io_magic, we start from DL1a (and in future we might also want to include DL1b).

From ctapipe v0.8.0 to v0.10.X, the organization of the containers changed. So I think the main task for the update would be to transform the DataContainer to the new structure. I haven't had much time to look into it, but I will start together with Jarred and the ctapipe_io_magic maintainers and keep you in the loop. Thanks!

Issue #106, opened 2021-12-10
Migrate to CTA DL1 data model v2.1.0


TjarkMiener:
After the release of ctapipe v0.12.0, the DL1 data model slightly changed.

TjarkMiener:
Solved by #107

PR #107, opened 2021-12-10
upgrade to ctapipe v0.12.0 (closses #106)

TjarkMiener:
@nietootein maybe it is worth to wait till we have a reference dataset with CTA data model v2.1.0 produced via ctapipe v0.12.0. Custom production can be processed using this PR.

### DIFF
--- a/dl1_data_handler/reader.py
+++ b/dl1_data_handler/reader.py
@@ -353,7 +353,7 @@ class DL1DataReaderSTAGE1(DL1DataReader):
         self.parameter_list = parameter_list
         self.image_channels = image_channels
         self.image_scale = None
-        self.peak_time_scale = None   
+        self.peak_time_scale = None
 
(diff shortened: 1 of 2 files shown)


TjarkMiener:
@nietootein Since the changes are really minor to the DL1 format, we should detect the data format version and support both versions, i.e. that we can read from prod5 (v1.0.0) and also from the latest ctapipe v0.12.0 (i.e. data model v2.1.0). When prod5 is outdated we can just remove the support for v1.0.0 by reverting [25f3805](https://github.com/cta-observatory/dl1-data-handler/pull/107/commits/25f3805151456e4a2764c679524dfd19b92ef59c).

nietootein (approved):
Go.

nietootein:
Merged. Thanks for the upgrade, @TjarkMiener!

PR #108, opened 2022-01-05
add class weights in reader

TjarkMiener:
Previously we calculated the class weights for unbalanced classes in CTLearn, which turned out to be very time consuming (iterating over the whole dataset). This commit will retrieve the number of events for each particle type on the fly in the init of the reader. Then one can calculate the class weights in CTLearn with the information.


nietootein (approved):
Go.

PR #110, opened 2022-01-13
moved pyirf

TjarkMiener:
It turned out to be unpractical to predict on the whole test sets at once, which makes the creation of the `SimulatedEventsInfo` inside the reader redundant.


nietootein (approved):
Go.

PR #111, opened 2022-01-19
bug fix - select tel id for stage1 datasets split by tel_type

TjarkMiener:
The select tel id option was broken, when processing stage1 datasets `split by tel_type`. However, this is not affecting any published results, i.e. the ICRC results, because stage1 datasets `split by tel_id` were used. It is fixed now but we probably want to only support `split by tel_id` in prediction mode in the future.

### DIFF
--- a/dl1_data_handler/reader.py
+++ b/dl1_data_handler/reader.py
@@ -446,23 +446,16 @@ class DL1DataReaderSTAGE1(DL1DataReader):
                 if "Subarray" not in multiplicity_selection:
                     multiplicity_selection["Subarray"] = subarray_multiplicity
 
+                # Construct the shower simulation table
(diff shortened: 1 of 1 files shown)


nietootein (approved):
Go.

dl1_data_handler/reader.py
```diff
@@ -489,20 +489,28 @@ def __init__(
...
-                        allevents = join(
-                            left=tel_type_table,
-                            right=simshower_table,
-                            keys=["obs_id", "event_id"],
-                        )
+                        tel_tables = []
+                        for tel_id in selected_telescopes[self.tel_type]:
+                            tel_table = tel_type_table[tel_type_table["tel_id"] == tel_id]
+
+                            simshower_table = read_table(
```
maxnoe, line 496: You don't need to read the shower table each time from the file again.

PR #112, opened 2022-02-01
code clean up

TjarkMiener:
- upgrade to MARS v2-20-1 superstar files
- aligned MAGIC files with ctapipe conventions to share the same CTLearn config (makes the migration to MAGIC stage1 files trivial in the future for CTLearn users)
- downgrade to python 3.7 again, because TF1.15 is not available for python 3.8 on pypi and CTLearn v0.6.0 with TF2.7 is not released yet.


nietootein (approved):
Go.

PR #117, opened 2022-12-22
Fix versioning of cta data models

TjarkMiener:
This is a minor bug fix about tracking the correct version of the CTA data model. At some point, we only want to support the latest one rather than all versions as it is for the time being. The PR also upgrades ctapipe to the latest version v0.17.


nietootein (approved):
Thanks so much, @TjarkMiener !

PR #119, opened 2023-05-10
R0 and R1 waveform reading for DL1DataReaderSTAGE1

TjarkMiener:
This PR adds the raw (R0) and calibrated (R1) waveform reading for the DL1DataReaderSTAGE1.

Calibrated (R1) waveform of a gamma shower through the ImageMapper:
waveform_gamma_imagemapper
Calibrated (R1) waveform of a proton shower through the ImageMapper:
waveform_proton_imagemapper
Raw (R0) waveform of a gamma shower through the ImageMapper:
rawwaveform_gamma_imagemapper

### DIFF
--- a/dl1_data_handler/reader.py
+++ b/dl1_data_handler/reader.py
@@ -43,7 +43,6 @@ lock = threading.Lock()
 
 class DL1DataReader:
     def __init__(self, file_list, mode="mono", subarray_info=None, event_info=None):
-
         # Construct dict of filename:file_handle pairs
         self.files = OrderedDict()
         # Order the file_list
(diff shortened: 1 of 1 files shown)


TjarkMiener:
The waveforms are provided in the "channel_last" format, i.e. the shape is (snapshots=40, image_dim_x, image_dim_y, channel=1). If 'IndexedConv'  is selected rather than a mapping method, the reader would return the vector in the shape (num_pixels, snapshots=40) as for the DL1 images where the shape is (num_pixels, image_channels={1,2}). Are those returning shapes fine for the GammaLearners @vuillaut @mikael10j?

nietootein (approved):
Excellent PR, @TjarkMiener ! Exciting experiments ahead ;)

PR #120, opened 2023-05-29
Bugfix transformation

TjarkMiener:
This PR is a rather quick fix. The az values are messed up for south-pointing datasets with a diffuse viewcone if the user set the north-pointing correction to true in the CTLearn config. With this bug fix, one can process south-pointing datasets correctly. Once I have more time, I will adopt the proper astropy function here and elsewhere in the code.


nietootein (approved):
Go!

PR #121, opened 2023-06-14
Fix error from unneeded memory use

aribrill:
Initializing a DL1DataReaderSTAGE1 on my machine results in a crash with the following error message:

MemoryError: Unable to allocate 23.4 GiB for an array with shape (262144, 120, 200) and data type float32

The cause is a few lines that are computing num_showers using shower histograms in the simulation files. In fact, num_showers is never subsequently used in reader.py (n_showers is used instead). Deleting these lines solves the error.

### DIFF
--- a/dl1_data_handler/reader.py
+++ b/dl1_data_handler/reader.py
@@ -268,12 +268,6 @@ class DL1DataReader:
                 n_showers = sum(np.array(runs.cols._f_col("n_showers"))) * shower_reuse
             else:
                 n_showers = sum(np.array(runs.cols._f_col("num_showers"))) * shower_reuse
-            if "service" in file.root.simulation:
-                service_table = file.root.simulation.service
-                shower_distributions = service_table._f_get_child("shower_distribution")
-                num_showers = np.sum(
-                    np.array(shower_distributions.cols._f_col("histogram"))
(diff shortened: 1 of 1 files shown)


TjarkMiener (approved):
Thanks @aribrill

maxnoe:
(262144, 120, 200) 

There is probably something wrong with your file, shower histograms are stored per simulation run. 

How do you have 262144 of them?

aribrill:
The files are these: https://zenodo.org/record/7298569

I don't know what's causing numpy to be allocating 262144 = 2^18 rows,
since there are only 50/200 obs_ids per file for the gammas/protons,
respectively.

maxnoe:
@aribrill 

Let me just comment here that you actually *should* look at the number of events in the histogram. At least for the prod5 and earlier simulations, there are quite a number of runs where the CORSIKA simulation stopped early, but we still have valid simtel output, just with less simulated showers than the configured number of showers in the run header.

So you will slightly overestimate the number of showers and thus under estimate e.g. effective area if you use the run header information instead of the histograms.

maxnoe:
This should be a bug in pytables. Is this deterministic? I.e. does this happen always for the same file?

PR #122, opened 2023-06-15
Fix telescope indexing assumption

aribrill:
In the CTAO dataset (https://zenodo.org/record/7298569), tel_ids do not follow strict sequential ordering. Of the 13 telescopes, the last two have tel_ids 35 and 19. In several places in DL1DataReader, the assumption is made that tel_ids are strictly sequential, which is incorrect in this case and breaks when loading MST data. This commit fixes this broken assumption by instead explicitly using the tel_ids in the order specified in the subarray layout table.


TjarkMiener (approved):
Thanks a lot for this bug fix, @aribrill!

PR #124, opened 2023-07-17
R0 pedsub

TjarkMiener:
This PR adds pixel-wise pedsub for R0. Also, the sequence selection to crop the waveforms is now done on the hexagonal 1D vector rather than the image itself, which is more efficient.

### DIFF
--- a/dl1_data_handler/reader.py
+++ b/dl1_data_handler/reader.py
@@ -466,6 +466,7 @@ class DL1DataReaderSTAGE1(DL1DataReader):
         waveform=None,
         waveform_sequence_length=None,
         waveform_format="timechannel_last",
+        waveform_r0pedsub=True,
         image_channels=None,
         mapping_settings=None,
         parameter_list=None,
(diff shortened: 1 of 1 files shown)


dl1_data_handler/reader.py
```diff
@@ -1333,60 +1337,70 @@ def _get_waveform(self, child, tel_type, waveform_index, img_child=None):
...
+                if "raw" in self.waveform:
+                    vector = vector[0]
+                waveform_max = np.argmax(np.sum(vector, axis=0))
             if dl1_cleaning_mask is not None:
-                cleaned_mapped_waveform = self.image_mapper.map_image(
-                    cleaned_vector, self._get_camera_type(tel_type)
-                )
+                waveform_max = np.argmax(np.sum(vector * dl1_cleaning_mask[:, None], axis=0))
+
+            # Retrieve the sequene around the shower maximum and calculate the pedestal
```
nietootein, line 1347: sequene -> sequence?

nietootein (commented):
Just a minor typo (see comment). All good otherwise!

nietootein (approved):
Go!

maxnoe:
Hi @TjarkMiener, @nietootein 

I am a bit confused why this is needed here (together with the R0 part of #119)  for a couple of reasons.

* R0 will never leave the telescopes, the first data level to be transmitted is the R1, which is at least roughly calibrated into pedestal subtracted photo electrons.
* The calibration you apply here is the same (hopefully) that is applied in the R0 to R1 step in the ctapipe SimTelEventSource, which we apply there so that ctapipe itself can start at R1 or DL0.

So why start from R0 here and re-implement the calibration?

TjarkMiener:
Hi @maxnoe,
For the adv. SiPM-based LST camera, people would like to explore the ability of CNN algorithms running on R0 inside the hardware of the camera (on FPGAs) for the trigger decision and later in the camera server to do gamma/hadron separation on R0 using stereoscopic information. Therefore, the support of R0 and R1 data reading is added here. I will have a look at the calibration of the ctapipe SimTelEventSource, however, mind that I'm only doing a pedestal subtraction to properly normalize the input image before passing it to the CNNs. I checked that I can reproduce the FADC_offset of the MCs.

maxnoe:
@TjarkMiener Oh, interesting! looking forward to seeing results then

PR #126, opened 2023-07-27
Data reading of the AI-based trigger system

TjarkMiener:
This PR suggests the data reading of the AI-based trigger system primarily designed for the SiPM-based LST cameras. It enables the ability to crop "trigger patches" with configurable positions and sizes, and retrieve the total number of Cherenkov photons (from the simulated true image) of that given patch. A fraction of NSB patches with no (or truncated) Cherenkov signal can be included by randomly selecting a trigger patch. With CTLearn, the user can then design a CNN-based model to reconstruct the number of Cherenkov photons in the trigger patch as a regression problem. This would allow us to make the trigger decision with different observation-dependent thresholds after the reconstruction.

### DIFF
--- a/dl1_data_handler/image_mapper.py
+++ b/dl1_data_handler/image_mapper.py
@@ -968,6 +968,21 @@ class ImageMapper:
             output_grid = np.column_stack([x_grid, y_grid])
 
         else:
+            if camera_type == "LSTCam":
+                # Additional smoothing of the 'y_ticks' array for LSTCam pixel positions.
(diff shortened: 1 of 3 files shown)


TjarkMiener:
Issue reported by Pablo Guzman (Big Thanks!) about the rotation and interpolation stripes were resolved with the latest commits!

nietootein (approved):
Go. Thanks, @TjarkMiener and @BastienLacave !

PR #128, opened 2024-05-31
Relative peak times

TjarkMiener:
We should work with relative peak times rather than absolute.


TjarkMiener:
Sorry, I was too fast earlier. Now it's working probably.  
Screenshot 2024-05-31 at 12 10 26

nietootein (approved):
Thanks, @TjarkMiener !

PR #130, opened 2024-06-03
Revert version.py

TjarkMiener:
Move back to previous version.py


nietootein (approved):

maxnoe:
I think you should probably just switch to using `setuptools-scm`

PR #133, opened 2024-06-14
Support only CTAO data format v6.0.0

TjarkMiener:
This PR upgrades to `ctapipe v0.21.1` and the CTAO data format `v6.0.0`. At the same time it drops support for older versions of the CTAO data format and the DL1DH data format. The deprecated writer functionality has also been removed.

In the future it would be good to work towards a reading API based on ctapipe component. Right now we have a lot of if else for different reading functionality such as stereo/mono, waveforms/images and trigger patches.


.github/workflows/python-package-conda.yml
```diff
@@ -25,7 +25,7 @@ jobs:
         echo $CONDA/bin >> $GITHUB_PATH
     - name: Install dependencies
       run: |
-        conda env update --file environment.yml --name base
+        conda env update --file environment.yml --name base --solver classic
```
TjarkMiener: I think `--solver classic` needs to be added to line 32, i.e. `conda install flake8 --solver classic`

nietootein (approved):
Go! Thanks, @TjarkMiener , @rcervinoucm !

nietootein:
We'll skip the Codacy issues for now...

PR #134, opened 2024-07-02
Allowing external algo to find trigger patch

TjarkMiener:
This PR allows to interface with an external algo (i.e. DBScan) to find the trigger patch. Information must be written into a `csv` file containing the `event_id`, `obs_id`, `tel_id`, `trg_pixel_id` (location; where the trigger happen via pixel_id), and `trg_waveform_sample_id` (time; at which waveform sample the trigger happen). The csv files have to be located at the same directory as the input data with the same naming convention ending with '`npe.csv`' instead of the standard `r0.dl1.h5`. The option to find the trigger patches from the true image or the integrated waveform is still possible.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -43,7 +43,7 @@ necessary package channels, and install dl1-data-handler specified version and i
 
 .. code-block:: bash
 
-   DL1DH_VER=0.11.1
+   DL1DH_VER=0.12.0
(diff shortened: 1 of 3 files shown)


nietootein (approved):
Go. Thanks again, @TjarkMiener !

PR #135, opened 2024-07-03
Use astropy for sky offset and separation calculation

TjarkMiener:
This PR enables the usage of astropy to get the sky offset and separation from tel pointing and source. We should in general re-think the way we are apply those transformation. However, this (bug) fix allows to process any arbitrary tel pointing in the sims.


nietootein (approved):
Go! Thanks, @TjarkMiener .

PR #139, opened 2024-07-15
example_identifiers_file is not able to open due a permissions issue.

Pablitinho:
example_identifiers_file fails to open due permissions issues.
For example:  

-rw-r--r-- 1 tjark.miener ctan-onsite-it 86051602 May 21 09:02 /fefs/aswg/workspace/tjark.miener/DeepCrab/logs/theta_16.087_az_108.090/LST1_cleanedTRN_type_fixcamgeo/example_identifiers_file.h5.

Opening in read mode I am able to open and read it without problems

### DIFF
--- a/dl1_data_handler/reader.py
+++ b/dl1_data_handler/reader.py
@@ -276,7 +276,7 @@ class DLDataReader:
         if example_identifiers_file is None:
             example_identifiers_file = {}
         else:
-            example_identifiers_file = pd.HDFStore(example_identifiers_file)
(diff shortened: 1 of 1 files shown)


TjarkMiener (approved):
Thanks @Pablitinho

TjarkMiener:
Sorry, yesterday I wasn't attentive enough to spot this. But, as discussed in private messages, the mode should not be specified here. If you are working on a cluster and you would like to use the checkpoints or trained models from other users hosted in folders with limited permission, simply cp the folder into your workspace and use it there. I reverted this in #140.

PR #141, opened 2024-07-16
Update release.yml

rcervinoucm:


maxnoe:
The version related stuff of the pypi step and the call to `setup.py` should also be changed.

Use `python -m build` instead.

Issue #104, opened 2021-09-01
Reduce RAM usage for reading ctapipe-stage1 files


TjarkMiener:
Following up on the discussion in [ctapipe PR 1730](https://github.com/cta-observatory/ctapipe/pull/1730).

I noticed a significant difference in the RAM usage during training time between the ctapipe-stage1 DL1 files, which are split by tel_ids and the one split by tel_types. 

The problem can be broken down to the lines [L792-L800](https://github.com/cta-observatory/dl1-data-handler/blob/stage1/dl1_data_handler/reader.py#L792-L800) and [L827-L837](https://github.com/cta-observatory/dl1-data-handler/blob/stage1/dl1_data_handler/reader.py#L827-L837) of the DL1DataReaderSTAGE1 in the [stage1 branch](https://github.com/cta-observatory/dl1-data-handler/blob/stage1). The core difference is that for the DL1 files split by tel_types, we only call `TABLE._f_get_child()` once in [L830](https://github.com/cta-observatory/dl1-data-handler/blob/stage1/dl1_data_handler/reader.py#L830), outside the for loop over the telescope ids, so only one `child` is in memory. My (probably poorly) pytables understanding here is that getting a `child` from a `node`  and then using `pytables indexing` is different than opening the whole table, storing the whole table temporarily in memory and retrieving the image by an index. @aribrill, can you please comment? 

As suggested by @maxnoe earsing the for loop when retreiving event images would be more efficient! I would need to check if multi-indexing on a `child` is supported by `pytables`. This feature is only applicable to the DL1 files split by tel_types, because for the DL1 files split by tel_ids a for loop is required because we have multiple `childs`/`tables`. The purpose of this function call (` __getitem__(self, idx)`) is to retrieve information (charges, hillas parameters, simulation properties etc.) for a single event for a short time (forward pass of the network). This function is called continually during the training process on the whole DL1 production to build batches of event images. The batches are shuffled in a sense that gamma and proton events from different runs (in particular different DL1 files) are contained within those batches.

maxnoe:
> for a single event for a short time (forward pass of the network).

I don't think you should read single events from file for that. Load tables with largish number of events that fit into RAM (configurable chunksize, like 10k to 50k events) and yield them one by one (without going back to the file) when the network needs a new batch. Read a new chunk when the current one is exhausted.

TjarkMiener:
[...]

PR #143, opened 2024-09-12
Define API for reader and image_mapper

### LINKED ISSUE #31: use more ctapipe functionality
I haven't been following the development of this code much, but looking quickly at the code, it seems to re-implement a lot of what is available in ctapipe, specifically the functionality of the HDF5TableWriter (which turns a ctapipe Container class into a row of a HDF5 table automatically). The ctapipe io system was designed exactly for the use case of writing DL1 data, so shouldn't be forgotten. We've been busy making it more useful recently as well, and there are still a few issues out for increased functionality.

I suspect much of this code could be simplified by using it (probably by >50%), and it gives you the advantage of keeping physical unit info, metadata, etc, which will be required in CTA output files. 

I'd suggest:
 * use `ctapipe.io.HDF5TableWriter` to simplify most of the tables you create, and get rid of your custom table definitions (the Container classes should be sufficient, and if not we can define new ones). You can basically throw away all of the code in `dump_event` this way. 
 * for the header info, I would suggest dumping the info from the `Provenance()` system, which contains all of the input files, output files, and the machine info, python version, etc. instead of using a custom solution. It gives you all this info automatically as a JSON object or dict. 
* I think you are misusing the EventSource classes - you never need to specify the subclass, just use a EventSourceFactory (or the `event_source()` helper function) , and it will construct the correct one based on the file contents (e.g. if you open a simtel file, it checks the magic number and return a SimTelEventSource), so the user just needs to specify a filename
* You mix your own argument parsing with the ctapipe Tool argument parsing, which is somewhat confusing. We should stick to one configuration system (though I agree the current one in ctapipe is not so friendly, we are working to make out own rather than use traitlets.config as we do now). If you use the ctapipe.core.Tool base class, you get provenance handling, and configuration for free.

TjarkMiener:
This PR suggests APIs for the reader and image_mapper. The configuration system and component scheme of `ctapipe` is adopted. Besides we adopt `astropy` tables for the batch generation. Reading is working in monoscopic and stereoscopic mode for DL1 images and R1 waveforms. The code properly process data with different array and telescope (divergent) pointings.  

A subclass designed for the adv. trigger system processing R0 waveforms will be added in a separated PR. 

–––
Closes #31 #104


dl1_data_handler/image_mapper.py
```diff
@@ -1,744 +1,1036 @@
...
+        # The default padding is removed after the conversion is finished.
+        self.internal_pad = 3
+
+        # Only needed for rebinnig
+        self.rebinning_mult_factor = 1
+
+        # Set the indexed matrix to None
+        self.index_matrix = None
+
+    def map_image(self, raw_vector):
```
Pablitinho, line 136: This is a general suggestion. Add the type in the function. **map_image**(self,raw_vector: np.array) (this is just an example). 
Reason: Easy to read, force the input type and also give you the chance of play with polymorphisms

dl1_data_handler/image_mapper.py
```diff
@@ -1,744 +1,1036 @@
...
-                    self.image_shapes[camtype][1] + self.default_pad * 2,
-                    self.image_shapes[camtype][2],
-                )
-            else:
-                self.image_shapes[camtype] = (
-                    self.image_shapes[camtype][0] + self.default_pad * 4,
-                    self.image_shapes[camtype][1] + self.default_pad * 4,
-                    self.image_shapes[camtype][2],
-                )
+    def _get_virtual_pixels(self, x_ticks, y_ticks, pix_x, pix_y):
```
Pablitinho, line 162: General suggestion: Add parameters meainig in "Doc Strings". You can use tools to generate it automatically in order to speed up this tedious task. Nobody like documentation 😂

dl1_data_handler/image_mapper.py
```diff
@@ -1,744 +1,1036 @@
...
+        """Create virtual hexagonal pixels outside of the camera."""
+        dist_first = np.around(abs(first_ticks[0] - first_ticks[1]), decimals=3)
+        dist_second = np.around(abs(second_ticks[0] - second_ticks[1]), decimals=3)
+
+        tick_diff = len(first_ticks) * 2 - len(second_ticks)
+        tick_diff_each_side = tick_diff // 2
+        # Extend second_ticks
+        for _ in range(tick_diff_each_side + self.internal_pad * 2):
+            second_ticks = (
+                [np.around(second_ticks[0] - dist_second, decimals=3)]
```
Pablitinho, line 182: Suggestion. Add decimals precision as a constant in order to remove MAGIC NUMBERS. 
Example: 

```
from collections import namedtuple
Constants = namedtuple('Constants', ['decimal_precision', 'tick_interval_limit'])
constants = Constants(3, 0.002)
print(constants.decimal_precision)
```

dl1_data_handler/image_mapper.py
```diff
@@ -1,744 +1,1036 @@
...
+        else:
+            x_grid, y_grid = np.meshgrid(grid_second, grid_first)
+        output_grid = np.column_stack([x_grid.ravel(), y_grid.ravel()])
+        return input_grid, output_grid
+
+    def _smooth_ticks(self, pix_pos, ticks):
+        """Smooth the ticks needed for the 'DigiCam' and 'CHEC' cameras."""
+        remove_val, change_val = [], []
+        for i in range(len(ticks) - 1):
+            if abs(ticks[i] - ticks[i + 1]) <= 0.002:
```
Pablitinho, line 339: 0.002 -> Magic number. Check the previous comment

dl1_data_handler/image_mapper.py
```diff
@@ -1,744 +1,1036 @@
...
+        self.geometry.rotate(self.geometry.pix_rotation)
+
+        self.pix_x = np.around(self.geometry.pix_x.value, decimals=3)
+        self.pix_y = np.around(self.geometry.pix_y.value, decimals=3)
+
+        self.x_ticks = np.unique(self.pix_x).tolist()
+        self.y_ticks = np.unique(self.pix_y).tolist()
+
+        # Additional smooth the ticks for 'DigiCam' and 'CHEC' cameras
+        if self.camera_type == "DigiCam":
```
Pablitinho, line 119: General suggestion: Is it not more convenient the use con Enum to avoid problems for camera_type or other potential types. 
Example: 

from enum import Enum
class Weekday(Enum):
    MONDAY = 1
    TUESDAY = 2
    WEDNESDAY = 3
    THURSDAY = 4
    FRIDAY = 5
    SATURDAY = 6
    SUNDAY = 7

print(Weekday.THURSDAY)

dl1_data_handler/reader.py
```diff
@@ -1,2028 +1,1202 @@
...
-        # Save the user attributes and useful information retrieved from the first file
-        self._v_attrs = self.files[first_file].root._v_attrs
-        self.subarray_layout = self.files[
-            first_file
-        ].root.configuration.instrument.subarray.layout
-        self.tel_ids = self.subarray_layout.cols._f_col("tel_id")
+        self.first_file = list(self.files)[0]
+        # Save the user attributes and useful information retrieved from the first file as a reference
+        self._v_attrs = self.files[self.first_file].root._v_attrs
         self.process_type = self._v_attrs["CTA PROCESS TYPE"]
```
Pablitinho, line 232: Here you can apply Enum for "CTA PROCESS TYPE" and others . More easy to change in the future if the name change.

dl1_data_handler/reader.py
```diff
@@ -1,2028 +1,1202 @@
...
-                            keys=["obs_id", "event_id"],
-                        )
+        # Handling the class weights calculation.
+        # Scaling by total/2 helps keep the loss to a similar magnitude.
+        # The sum of the weights of all examples stays the same.
+        self.class_weight = None
+        if self.process_type == "Simulation":
+            if self.bkg_input_files is not None:
+                self.class_weight = {
+                    0: (1 / self.n_bkg_events) * (self._get_n_events() / 2.0),
```
Pablitinho, line 360: 1.0/self.n_bkg_events

Pablitinho (commented):
First revision round 🙂

dl1_data_handler/image_mapper.py
```diff
@@ -1,744 +1,1036 @@
...
+            "LSTSiPMCam": 234,
+            "FlashCam": 112,
+            "NectarCam": 110,
+            "SCTCam": 120,
+            "DigiCam": 96,
+            "CHEC": 48,
+            "ASTRICam": 56,
+            "VERITAS": 54,
+            "MAGICCam": 78,
+            "FACT": 92,
```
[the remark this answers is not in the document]
TjarkMiener: Yes that is on purpose. I needed to add two more. Probably because we didn't de-rotated the pixel positions by default before. Those default images shapes are kind of magic numbers originated from the oversampling method, where one hexagonal pixels is divided into 4 squared pixels. In principle, those numbers can be retrieved from the `_get_grid()` function of the `OversamplingMapper` for any `ImageMapper` subclass.

dl1_data_handler/image_mapper.py
TjarkMiener: I did not add the parameters for the internal functions of the `ImageMapper` class, since I thought it was not necessary. Only for the one (`map_image()`) that can be called  from outside. If you are insisting, I can add them here as well ;-)

dl1_data_handler/image_mapper.py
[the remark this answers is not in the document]
TjarkMiener: Solved

Pablitinho (approved):
LGTM

dl1_data_handler/image_mapper.py
```diff
@@ -1,744 +1,1112 @@
...
+        self.geometry.rotate(self.geometry.pix_rotation)
+
+        self.pix_x = np.around(
+            self.geometry.pix_x.value, decimals=constants.decimal_precision
+        )
+        self.pix_y = np.around(
+            self.geometry.pix_y.value, decimals=constants.decimal_precision
+        )
+
+        self.x_ticks = np.unique(self.pix_x).tolist()
```
mexanick, line 103: Am I right you assume regularly spaced pixels for any kind of camera geometry? If not, did you test any geometry with shuffle step (e.g. square pixels where rows are shifted by e.g. 25%?)

dl1_data_handler/image_mapper.py
```diff
@@ -1,744 +1,1112 @@
...
+        self.y_ticks = np.unique(self.pix_y).tolist()
+
+        # Additional smooth the ticks for 'DigiCam' and 'CHEC' cameras
+        if self.camera_type == "DigiCam":
+            self.pix_y, self.y_ticks = self._smooth_ticks(self.pix_y, self.y_ticks)
+        if self.camera_type == "CHEC":
+            self.pix_x, self.x_ticks = self._smooth_ticks(self.pix_x, self.x_ticks)
+            self.pix_y, self.y_ticks = self._smooth_ticks(self.pix_y, self.y_ticks)
+
+        # At the edges of the cameras some mapping methods run into issues.
```
mexanick, line 113: Is this because your "ticks" maxes out at the maxima of pix_x, pix_y and do not take into account the pixel's area (border)?

dl1_data_handler/image_mapper.py
```diff
@@ -1,744 +1,1112 @@
...
+        gridpoints = np.array(np.meshgrid(x_ticks, y_ticks)).T.reshape(-1, 2)
+        gridpoints = [tuple(l) for l in gridpoints.tolist()]
+        virtual_pixels = set(gridpoints) - set(zip(pix_x, pix_y))
+        virtual_pixels = np.array(list(virtual_pixels))
+        return virtual_pixels
+
+    def _create_virtual_hex_pixels(
+        self, first_ticks, second_ticks, first_pos, second_pos
+    ):
+        """Create virtual hexagonal pixels outside of the camera."""
```
mexanick, line 165: (even inline) will help

dl1_data_handler/image_mapper.py
```diff
@@ -1,744 +1,1112 @@
...
+        )
+        if len(self.x_ticks) < len(self.y_ticks):
+            x_grid, y_grid = np.meshgrid(grid_first, grid_second)
+        else:
+            x_grid, y_grid = np.meshgrid(grid_second, grid_first)
+        output_grid = np.column_stack([x_grid.ravel(), y_grid.ravel()])
+        return input_grid, output_grid
+
+    def _smooth_ticks(self, pix_pos, ticks):
+        """Smooth the ticks needed for the 'DigiCam' and 'CHEC' cameras."""
```
mexanick, line 426: Why this is needed?

dl1_data_handler/image_mapper.py
```diff
@@ -1,744 +1,1112 @@
...
+        self.internal_shape = self.image_shape
+
+        # Create square grid
+        input_grid, output_grid = self._get_square_grid()
+        # Calculate the mapping table
+        self.mapping_table = super()._generate_nearestneighbor_table(
+            input_grid, output_grid, pixel_weight=1.0
+        )
+
+    def _get_square_grid(
```
mexanick, line 481: Looks like this method has similar meaning to `get_grids()` of the other derived classes, or `get_grids_for_interpolation()` of the parent class, but has a different name. Consider a uniform name and an abstract method (or some sort of default implementation) in the base class.

dl1_data_handler/image_mapper.py
```diff
@@ -1,744 +1,1112 @@
...
+        **kwargs,
+    ):
+        super().__init__(
+            geometry=geometry,
+            config=config,
+            parent=parent,
+            **kwargs,
+        )
+
+        if geometry.pix_type != PixelShape.HEXAGON:
```
mexanick, line 784: why one can't oversample a square pixel grid?

dl1_data_handler/image_mapper.py
```diff
@@ -1,744 +1,1112 @@
...
```
mexanick, line 889: same question as above

dl1_data_handler/reader.py
```diff
@@ -1,2028 +1,1206 @@
...
+
+    This class handles the initialization and configuration of the data reader, including setting up quality criteria,
+    managing input files, and extracting relevant information from the data files. It supports both observational and
+    simulation data, and can operate in ``mono`` and ``stereo`` modes.
+
+    Attributes
+    ----------
+    quality_query : TableQualityQuery
+        An instance of TableQualityQuery to apply quality criteria to the data.
+    files : OrderedDict
```
mexanick, line 79: Why do you need an OrderedDict here? I don't see any use of specific features for it in the code. On the other hand, a standard dict since python 3.6 retains original order of items.

dl1_data_handler/reader.py
```diff
@@ -1,2028 +1,1206 @@
...
-                    {
-                        "name": "parameters",
-                        "tel_type": self.tel_type,
-                        "base_name": "parameters",
-                        "shape": ((1,) + (len(self.parameter_list),)),
-                        "dtype": np.dtype(np.float32),
-                    }
-                )
+                def _multiplicity_cut_tel_type(table, key_colnames):
+                    self.min_telescopes_of_type.attach_subarray(self.subarray)
```
mexanick, line 564: bad design, you modify here some objects that are out of scope for this function. Also you don't use `key_colnames` local variable.

dl1_data_handler/reader.py
```diff
@@ -1,2028 +1,1206 @@
...
-                        }
-                    ]
+                table_per_type.add_column(tel_type_id, name="tel_type_id", index=3)
+                events.append(table_per_type)
+            events = vstack(events)
+            # Apply the multiplicity cut based on the subarray
+            events = events.group_by(["obs_id", "event_id"])
+
+            def _multiplicity_cut_subarray(table, key_colnames):
+                return len(table) >= self.min_telescopes
```
mexanick, line 578: why do you need `key_colnames`?

mexanick (commented):
A few comments inline. It is quite difficult to comment on such a huge PR. 
I suggest an implementation of a proper testing suite instead of a test notebook in order to automatize testing. Also, would be great to see a class diagram for the refactored ImageMapper showing what functionality is inherited, what is overloaded and what is extended.

dl1_data_handler/reader.py
TjarkMiener: `Astropy` filter functionality requires a function with those two arguments. See [here](https://docs.astropy.org/en/stable/table/operations.html#filtering).

TjarkMiener:
I'm done with my last fixes which I wanted to include here @Pablitinho @mexanick. In order not to populate even more this PR (Sorry!), I'm suggesting to merge after a second look. We can fix with smaller dedicated PRs if needed. @nietootein would that be fine with you?

dl1_data_handler/image_mapper.py
```diff
@@ -1,744 +1,1112 @@
...
+    "NearestNeighborMapper",
+    "OversamplingMapper",
+    "RebinMapper",
+    "ShiftingMapper",
+    "SquareMapper",
+]
+
+# Constants for the ImageMapper classes
+Constants = namedtuple("Constants", ["decimal_precision", "tick_interval_limit"])
+constants = Constants(3, 0.002)
```
mexanick, line 28: Make it an ImageMapper class attribute (in this way, all objects will have the same value, but you won't have the global var.

dl1_data_handler/image_mapper.py
```diff
@@ -1,744 +1,1112 @@
-import numpy as np
-import logging
-import bisect
+"""
+This module defines the ``ImageMapper`` classes, which holds the basic functionality for mapping raw 1D vectors into 2D mapped images.
```
mexanick, line 2: which holds --> that hold

README.rst
```diff
@@ -60,7 +60,7 @@ The main dependencies are:
 
 * PyTables >= 3.8
 * NumPy >= 1.20.0
-* ctapipe == 0.21.2
+* ctapipe == 0.22.0
```
mexanick, line 63: Doesn't correspond to env file (>=0.23 there). Consider automating (for the future PR!)

dl1_data_handler/reader.py
```diff
@@ -1,2028 +1,1635 @@
-from collections import Counter, OrderedDict
-import random
-import threading
+"""
+This module defines the ``DLDataReader`` and ``TableQualityQuery`` classes, which holds the basic reading and processing functionality for Deep Learning (DL) analyses.
```
mexanick, line 2: which holds --> that hold

dl1_data_handler/image_mapper.py
```diff
@@ -1,744 +1,1112 @@
...
-                try:
-                    from ctapipe.instrument.camera import CameraGeometry
-                except ImportError:
-                    raise ImportError(
-                        "The `ctapipe.instrument.camera` python module is required, if pixel_positions is `None`."
+        self.geometry = geometry
+        self.camera_type = self.geometry.name
+        self.n_pixels = self.geometry.n_pixels
+        # Rotate the pixel positions by the pixel to align
+        self.geometry.rotate(self.geometry.pix_rotation)
```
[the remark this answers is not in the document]
maxnoe: I think this aligns the pixel grid to the coordinate axis.

Pablitinho:
setup.cfg has a deprecated line.
Now: 

[tool:pytest]
pep8ignore =

Action (remove pep8ignore line): 
[tool:pytest]

If you dont remove this line you will have a warning in pytest like this: 

`============================================================================================================================ warnings summary =============================================================================================================================
../../../miniconda3/envs/ctlearn/lib/python3.11/site-packages/_pytest/config/__init__.py:1441
  /Users/pabloguzmansanchez/miniconda3/envs/ctlearn/lib/python3.11/site-packages/_pytest/config/__init__.py:1441: PytestConfigWarning: Unknown config option: pep8ignore
  
    self._warn_or_fail_if_strict(f"Unknown config option: {key}\n")

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html`

After remove the warning is out.

TjarkMiener:
I piggybacked on ctapipe for the test data in order to implement a minimal unit test. Can you please check the latest commit @Pablitinho thanks!

TjarkMiener:
@Pablitinho thanks for noticing, I just moved the loader class to ctlearn because of the keras/tensorflow dependency. This is solving the tests. I will merge #142 and rebase this branch.

TjarkMiener:
Thanks a lot @rcervinoucm. CI is passing now with 'minimal' meaningful unit testing. This PR also includes the preparation for dl1dh v0.13.0. Anything missing or can we go ahead? @Pablitinho @nietootein @rcervinoucm

Pablitinho:
@TjarkMiener 

**UnclosedFileWarning:**

`/Users/pp/miniconda3/envs/ctlearn/lib/python3.11/site-packages/tables/file.py:114: UnclosedFileWarning: Closing remaining open file: /private/var/folders/qk/qydpv_9n4b5g2pl834gyrcw80000gn/T/pytest-of-pabloguzmansanchez/pytest-31/r1_0/gamma.r1.h5
  warnings.warn(UnclosedFileWarning(msg))
/Users/pp/miniconda3/envs/ctlearn/lib/python3.11/site-packages/tables/file.py:114: UnclosedFileWarning: Closing remaining open file: /private/var/folders/qk/qydpv_9n4b5g2pl834gyrcw80000gn/T/pytest-of-pabloguzmansanchez/pytest-31/dl1_0/gamma.dl1.h5
  warnings.warn(UnclosedFileWarning(msg))` 

In order to avoid this problem (the files are not closed), the most appropiate solution is the next:


`
...
import atexit 

...
class DLDataReader(Component):

    def __init__(
        self,
        input_url_signal,
        input_url_background=[],
        config=None,
        parent=None,
        **kwargs,
    ):
         ....
         atexit.register(self.__destructor)
          ....

    def __destructor(self):
        """Destructor to ensure all opened HDF5 files are properly closed."""
        if hasattr(self, "files"):  # Ensure self.files exists before attempting to close
            for file_name in list(self.files.keys()):
                try:
                    if self.files[file_name].isopen:  # Check if file is still open
                        print(f"Closing file: {file_name}")
                        self.files[file_name].close()
                except Exception as e:
                    print(f"Error closing file {file_name}: {e}")

`
Dont use functions like **__del__** or **__exit__** . del is called when the user **del** the object and **__exit__** is when is out of the scope.

This is a suggestion, apply it if you wish.

PR #150, opened 2025-02-24
Adopt ctapipe coordinate frames

### LINKED ISSUE #145: Clarify geometry rotation
What is the purpose of this? If I understand correctly, `pix_rotation` is an angle, at which every pixel is rotated, but not necessarily the camera, as for that one there's `cam_rotation`. From the comment above this line, perhaps `cam_rotation` shall be used instead?

_Originally posted by @mexanick in https://github.com/cta-observatory/dl1-data-handler/pull/143#discussion_r1772933996_

TjarkMiener:
This PR adopts the handling of coordinate frames from `ctapipe`. Labeling this as a bug fix since those changes are really necessary for the correct transformation back to alt az coordinates and for properly handling stereo data.

Closes #145

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -21,7 +21,7 @@ DL1 Data Handler
     :target: https://github.com/cta-observatory/dl1-data-handler/actions/workflows/python-package-conda.yml
     :alt: Continuos Integration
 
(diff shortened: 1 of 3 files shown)


nietootein (approved):
Go! Thanks, @TjarkMiener.

PR #151, opened 2025-03-10
DLRawTriggerReader

### LINKED ISSUE #148: Enable reading R0 data from ctapipe processor
Reading from R0/R1 is currently missing and is needed for those projects that take waveforms as input.

jbuces:
This PR allows the possibility to process r0 waveforms, and multiple forms of treating trigger patches of these raw waveforms. 
Fixes #148 .
[DLRawTriggerReader.pdf](https://github.com/user-attachments/files/19171379/DLRawTriggerReader.pdf)


dl1_data_handler/image_mapper.py
```diff
@@ -91,6 +97,8 @@ def __init__(
         self.camera_type = self.geometry.name
         self.n_pixels = self.geometry.n_pixels
         # Rotate the pixel positions by the pixel to align
+        if self.camera_type == "UNKNOWN-7987PX":
+            self.geometry.pix_rotation = 8.213 * u.deg
```
TjarkMiener: Why this pixel rotation have to be hardcoded and is not in the cam geometry object from the file?

dl1_data_handler/image_mapper.py
```diff
@@ -91,6 +97,8 @@ def __init__(
         self.camera_type = self.geometry.name
         self.n_pixels = self.geometry.n_pixels
         # Rotate the pixel positions by the pixel to align
+        if self.camera_type == "UNKNOWN-7987PX":
```
TjarkMiener: Can you maybe overwrite the name of the geometry after reading in the camera geometry object from the file, i.e. in the reader?

dl1_data_handler/image_mapper.py
```diff
@@ -648,6 +657,89 @@ def _get_grids(
...
+            parent=parent,
+            **kwargs,
+        )
+
+        if geometry.pix_type != PixelShape.HEXAGON:
+            raise ValueError(
+                f"HexagonalPatchMapper is only available for hexagonal pixel cameras. Pixel type of the selected camera is '{geometry.pix_type}'."
+             )
+
+        path = files("dl1_data_handler.ressources").joinpath("newcam_sipm_patches.h5")
```
TjarkMiener: This file needs a versioning and a proper naming structure. I'm proposing 
triggerpatches_AdvCam_v1.h5

dl1_data_handler/image_mapper.py
```diff
@@ -648,6 +657,89 @@ def _get_grids(
...
+
+        if geometry.name == "UNKNOWN-7987PX":
```
TjarkMiener: the line from above should only be executed once you the AdvCam selected, i.e. move it here.

dl1_data_handler/image_mapper.py
TjarkMiener: If you rename "`UNKNOWN-7987PX`" earlier to AdvCam, you could retrieve this file via `f"{triggerpatches_{geometry.name}_v1}"`. Also the version should be a configuration. At this stage of the project we might want to experiment with different patches, so we need to be able to switch between different version without reinstalling the code.

dl1_data_handler/reader.py
```diff
@@ -242,6 +245,13 @@ class DLDataReader(Component):
         ),
     ).tag(config=True)
 
+    quality_cuts = Bool(
+        default_value=True,
+        help=(
+            "Require quality cuts by ``TableQualityQuery``. False for nsb trigger."
+        ),
+    ).tag(config=True)
+
```
TjarkMiener: `TableQualityQuery` should be empty if you do not want to pass any quality cuts. I know it is a quite convenient shortcut for you, but I would prefer to have properly handled by the configuration system. We should write example configs for the different use cases.

dl1_data_handler/reader.py
```diff
@@ -255,7 +265,10 @@ def __init__(
         # Register the destructor to close all open files properly
         atexit.register(self.__destructor)
         # Initialize the Table data quality query
-        self.quality_query = TableQualityQuery(parent=self)
```
TjarkMiener: See comment above

dl1_data_handler/reader.py
```diff
@@ -1180,6 +1220,14 @@ def __destructor(self):
     def _append_features(self, batch) -> Table:
         pass
 
+    @abstractmethod
+    def _get_raw_example(self, batch) -> Table:
+        pass
+
+    @abstractmethod
+    def _add_tdscan_table(self,batch) -> Table:
+        pass
```
TjarkMiener: Why are they `abstractmethod` if they are only used in the trigger subclass? I think they should be removed here and elsewhere (except in the trigger subclass).

dl1_data_handler/reader.py
```diff
@@ -1574,9 +1641,10 @@ def __init__(
         )
 
         # Read the readout length from the first file
+        data_group = getattr(self.files[self.first_file].root, self.data_level)
```
TjarkMiener: why here we have `self.data_level` and above `data_level = "r1"`? I think each reader subclass should hold a datalevel attribute from the enum (see comment above). Then you can your desired string `self.data_level.name.lower()`.

dl1_data_handler/reader.py
```diff
@@ -1697,6 +1774,564 @@ def _append_features(self, batch) -> Table:
...
+            "``double_patches``: extract the sequence of selected samples for a random nsb patch and the patch containing the brightest pixel. "
+            "``max_patches``: extract all the nsb patches and all the patches above the cpe_threshold. "
+        ),
+    ).tag(config=True)
+
+    output_size = CaselessStrEnum(
+        ["patch", "sector", "camera"],
+        default_value = "camera",
+        help=(
+            "Set the number of pixels in the output vector, patch and sector only available for the Advanced SiPM camera. "
```
TjarkMiener: for the Advanced SiPM camera. -> for the Advanced SiPM camera (AdvCam).

dl1_data_handler/reader.py
```diff
@@ -1697,6 +1774,564 @@ def _append_features(self, batch) -> Table:
...
+            "``patch``: 343 pixels trigger patches."
+            "``sector``: 29898 pixels patches, approx. 1/3 of the camera. "
+            "``camera``: complete camera. "
+        ),
+    ).tag(config=True) 
+
+    add_dummy_channel = Bool(
+        default_value = False,
+        allow_none = False,
+        help=("Boolean variable to add or not an extra dummy dimension for the channel")
```
TjarkMiener: Please add a little bit more information for this help. E.g. why this is needed/when the user needs to do this action oon swiitching the boolean on.

dl1_data_handler/reader.py
```diff
@@ -1697,6 +1774,564 @@ def _append_features(self, batch) -> Table:
...
+    ).tag(config=True) 
+
+    subtract_mean = Int(
+        default_value = None,
+        allow_none = True,
+        help=("Value to subtract to the waveform")
```
TjarkMiener: Add more information here please. From where you calculate the mean? Do you calculate for each pixels or over the whole camera? Are you using pedestal events for this?

notebooks/test_r0_reader.ipynb
TjarkMiener: Thanks for putting together this detailed notebook. Can you please tell me the file size of e.g. `/lustre/ific.uv.es/ml/ucm147/datasets/TriggerlessData/gammas/corsika_run110.r0.h5`? We can consider uploading it to the test data and run it in the CI as unittests.

TjarkMiener (changes requested):
Thanks so much @jbuces for all the work! Overall it looks already good. I sent some comments. Also can you do a proper rebase to the `main` please?

TjarkMiener:
@jbuces can you please rebase to the current main. We pushed some changes to the CI. Thanks in advance!

dl1_data_handler/image_mapper.py
```diff
@@ -664,6 +670,99 @@ def _get_grids(
         return input_grid, output_grid
 
 
+class HexagonalPatchMapper(ImageMapper):
```
tjarkmiener, line 673: @jbuces you are breaking the `ImageMapper` API with `HexagonalPatchMapper`, i.e. calling `map_image(raw_vector)` results in:

```
   images = np.concatenate(
            [
               (raw_vector[:, channel].T @ self.mapping_table).reshape(
                                            ^^^^^^^^^^^^^^^^^^
                    self.image_shape, self.image_shape, 1
                )
                for channel in range(raw_vector.shape[1])
            ],
            axis=-1,
        )
E       AttributeError: 'HexagonalPatchMapper' object has no attribute 'mapping_table'

dl1_data_handler/image_mapper.py:167: AttributeError
```

TjarkMiener (changes requested):

dl1_data_handler/image_mapper.py
jbuces: For this I created the cam_neighbor_array variable initialised in ImageMapper to None, and we give a value when HexagonalPatchMapper is configured, so that the reader never calls map_image when the HexagonalPatchMapper is called. But yes it breaks, what would be the solution? Generating a dummy mapping_table? Since the HexagonalPatchMapper should never return a 2d image

tjarkmiener: Not sure at the moment. But if the `HexagonalPatchMapper` is not solving the same problem as the `ImageMapper`, it should probably not inherit from it. Are you using any functionality from the `ImageMapper` parent?

tjarkmiener: You could have a PatchMapper API that is only for the TriggerReader. Would this simplify things in the reader classes where you check for `cam_neighbor_array`?

jbuces: I am using it to retrieve the number of pixels (for unmapped shape), and the geometry for passing it to get_unmapped_waveform but this function only uses it to clean the waveform with DBSAN (still not implemented). For the rest I only call it for characteristics that I retrieve in the HexagonalPatchMapper and deciding whether or not leaving the waveform or image unmaped with this cam_neighbor_array

PR #152, opened 2025-03-11
Bugfix pixrotation

TjarkMiener:
This is a bugfix for inserting the correct pixel rotation in the camera frame.


nietootein (approved):
Go! Thanks, @TjarkMiener!

PR #155, opened 2025-03-28
Add choice of focal length

TjarkMiener:


dl1_data_handler/reader.py
```diff
@@ -192,6 +194,21 @@ class DLDataReader(Component):
         ),
     ).tag(config=True)
 
+    focal_length_choice = UseEnum(
+        FocalLengthKind,
+        default_value=FocalLengthKind.EQUIVALENT,
```
maxnoe, line 199: Please don't use `EQUIVALENT` as default. It will result in biased reconstructions. Use `EFFECTIVE` and let people explicitly make the choice of going back to `EQUIVALENT` for very old simulations files that do not have the `EFFECTIVE` information.

maxnoe: See: https://github.com/cta-observatory/ctapipe/issues/1908

TjarkMiener: Thanks @maxnoe I set it to `EQUIVALENT` as default because I have asked Yorgos who is producing the LST-1 simulations. And in his `simtel` configuration file the `EFFECTIVE` focal length is set but there is a comment saying that simtel will ignore this value. So I thought in order to transform the simulated arrival direction from AltAz to derotated camera frame, I would need to use the `EQUIVALENT` focal length since this is the one used in `simtel`. All the reconstruction happen in the derotated camera frame and then we transform back using the `EFFECTIVE` focal length for real LST-1 data in `CTLearn`. Do we need to use `EFFECTIVE` focal length consistently in this case?

maxnoe: Effective is ignored in the sense that it is not needed by sim_tellarray since it is using real ray tracing of the actual reflector and the effective focal length is a computed quantity from that.

In the analysis, please always use the effective focal length if it is available.

If you transform sky positions to camera using the equivalent focal length, the result will be systematically off, especially off center.

TjarkMiener: Okay thanks for the clarification!

nietootein: Thanks for the clarification, @maxnoe. Much appreciated!

nietootein (approved):
Thanks, @TjarkMiener !

PR #160, opened 2025-04-03
Update python version according to SPEC0 (include 3.13, drop 3.10)

rcervinoucm:


TjarkMiener (approved):
Thanks a lot @rcervinoucm

PR #164, opened 2025-05-15
Fix duplicates

TjarkMiener:
Sometimes (very rarely) simtel writes duplication of events (maybe slurm glitches but it is unknown, observed with same LST sims discussed in [ctapipe#2344](https://github.com/cta-observatory/ctapipe/issues/2344), but also for regular LST-1 sims). Normally, it is caught by diagnostic scripts. However, if there are just a few duplications the script may pass. Therefore, we have problems later on. The dl1_lookup flag is for repairing existing simtel productions.

### DIFF
--- a/dl1_data_handler/reader.py
+++ b/dl1_data_handler/reader.py
@@ -536,6 +536,8 @@ class DLDataReader(Component):
                     dl1_tel_table.add_column(
                         np.arange(len(dl1_tel_table)), name="table_index", index=0
                     )
+                    # Unique the table to remove unwanted duplication 
+                    dl1_tel_table = unique(dl1_tel_table, keys=["obs_id", "event_id", "tel_id"])
                     # Join the DL1 image table with the DL1 parameter table
                     tel_table = join(
                         left=tel_table,
@@ -673,6 +675,8 @@ class DLDataReader(Component):
                         dl1_tel_table.add_column(
                             np.arange(len(dl1_tel_table)), name="table_index", index=0
                         )
+                        # Unique the table to remove unwanted duplication
+                        dl1_tel_table = unique(dl1_tel_table, keys=["obs_id", "event_id", "tel_id"])
                         # Join the DL1 image table with the DL1 parameter table
                         tel_table = join(
                             left=tel_table,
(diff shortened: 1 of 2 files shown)


nietootein (approved):
Go! Many thanks for all the detective work here, @miener!

maxnoe:
Duplicate events? This is the first time hearing of this and this is quite concerning. You should certainly report to @kbernloehr.

Do you an example file at hand?

dl1_data_handler/reader.py
```diff
@@ -536,6 +536,8 @@ def _construct_mono_example_identifiers(self):
                     dl1_tel_table.add_column(
                         np.arange(len(dl1_tel_table)), name="table_index", index=0
                     )
+                    # Unique the table to remove unwanted duplication 
+                    dl1_tel_table = unique(dl1_tel_table, keys=["obs_id", "event_id", "tel_id"])
```
maxnoe, line 540: I think this should not happen silently but raise a massive warning

maxnoe: Also note that `unique` sorts the output table  by `keys`. This can be an unwanted side effect.

TjarkMiener:
> Do you an example file at hand?

I will prepare a proper report of this and provide example files to reproduce (properly not before the GM next week).

dl1_data_handler/reader.py
TjarkMiener: Yes, some warnings would be good here. Also, I think the `process-tool` should not let me write in one table two entries with equal keys (obs, event, and tel ids). Not sure if this would be easy to check?

Thanks for the heads-up for the sorting. In the past, some ordering or not ordering of astropy functions gave me some headache, but here we do not care about the order since the table index for the row is added as a column before. `astropy v7.X ` introduces some nice functionality to `keep_order` (but I think you are well aware for it ;-) )

TjarkMiener:
I'm not 100% sure, but I think I find a suitable example file. Path in the lst-cluster is `/fefs/aswg/mc/LST_Advanced_Camera_Prod1/gamma-diffuse/zenith_20deg/south_pointing/4LSTs_SiPM/sim_telarray/simtel_corsika_run9.simtel.gz`. The event with obs_id=9, event_id=3986307, tel_id=4, should be duplicated. @maxnoe

PR #166, opened 2025-08-21
Add additional simulated parameters to the batch reading 

TjarkMiener:
This PR adds the impact h_max and first interaction to read in the batch.


TjarkMiener:
@burmist-git Should we add a "transformation" for the radius of the impact? See e.g. [energy & direction transformation](https://github.com/cta-observatory/dl1-data-handler/blob/main/dl1_data_handler/reader.py#L591-L593).

PR #176, opened 2025-12-15
fix release workflows

TjarkMiener:


TjarkMiener:
@rcervinoucm @nietootein FYI I sneaked in two commits relaxing the subarray equal check; this is needed to properly test the classification task [ctlearn-14484ca](https://github.com/ctlearn-project/ctlearn/pull/269/commits/14484ca28c419212efa7edab7ea52ece6be418ee). Disabling the strict check for the subarrays is needed to make the unittests pass with the current test data from the different sites of each particle type.

PR #181, opened 2026-02-12
allow ctapipe v0.25 and test minimal ctapipe version

TjarkMiener:


vuillaut:
Thanks @TjarkMiener 
Just FYI I don't think meta.yml is used for the package on conda-forge, but maybe it is for the ctlearn-project channel ?

TjarkMiener:
Actually, it is also needed for the `manager` @BastienLacave @nietootein @rcervinoucm

PR #183, opened 2026-03-13
remove allow_failure for python 3.14

TjarkMiener:
pytables 3.11.X has been released which supports python 3.14. So we can remove the allow_failure for this version


rcervinoucm (approved):
Thanks a lot @TjarkMiener!

PR #185, opened 2026-05-05
bump pytables to v3.11 to support python v3.14

TjarkMiener:


rcervinoucm (approved):
Thanks a lot @TjarkMiener!

PR #187, opened 2026-07-17
vectorized map_image() function and support batched imagesto be mapped

tjarkmiener:


alexcervino (approved):
Thanks a lot @TjarkMiener , very useful enhancement

PR #192, opened 2026-08-28
Add HexagdlyMapper for hex-native grid mapping

YugnatD:
## Summary

Adds `HexagdlyMapper`, an `ImageMapper` that maps a hexagonal camera image onto the exact offset-column grid `keras-hexagdly`'s hexagonal convolution layers expect, instead of interpolating onto a square grid like the existing mappers do. Split out of ctlearn-project/ctlearn#286 per @tjarkmiener's review there.

## Test plan
- [x] New `TestHexagdlyMapperSpecific` class in `test_image_mapper.py`: zero-neighbour-mismatch check against LSTCam/MAGICCam/NectarCam/FlashCam/DigiCam, square-pixel-camera rejection, exact pixel placement, batch/multichannel support (existing generic mapper tests extended to cover it too), a synthetic-chirality-flip test, and a test confirming the origin search runs once at construction, not per image
- [x] Full `test_image_mapper.py` suite passes (58/58)


vuillaut:
Hi. 
I'm confused, isn't it exactly what IndexedConv do ?

YugnatD:
> Hi. I'm confused, isn't it exactly what IndexedConv do ?

Hi,

Same goal, yes both avoid resampling hexagonal data onto a square grid.

The difference is the mechanism: HexagDLy ([https://arxiv.org/pdf/1903.01814](https://arxiv.org/pdf/1903.01814)) uses
an offset-column addressing where the hexagonal neighbourhood is exactly
representable in 2D, and decomposes the hex kernel into rectangular sub-kernels,
so standard convolutions do the work. This PR is just the mapper producing that
addressing, with the neighbour topology verified exactly.

The practical gain is memory: nothing extra gets materialised at each layer, so
memory barely grows when you use larger kernels, which can allow a bigger batch
size. The cost is speed: HexagDLy is slower per step.

I'm also working on the hls4ml support, and we're currently running some tests on
it for SST-1M. (For now I consider that only the io_stream mode for Conv2d and
MaxPool2d is working; I still need to think about how to do it in 3D, though if
I'm not wrong, even the standard Conv3D layer isn't supported in hls4ml yet.)

Another advantage is the new share_neighbors parameter on Conv2d and Conv3d,
which could help a model learn symmetrical/diagonal/ring patterns while reducing the number of
weights and keeping the same performance, this could help for an FPGA
implementation, even though the number of operations stays the same.

As it's the direction we took for SST-1M, I was wondering if it could be useful for CTA.

Not proposed as a replacement for IndexedConv, just another no-resampling route.

vuillaut:
> > Hi. I'm confused, isn't it exactly what IndexedConv do ?
> 
> Hi,
> 
> Same goal, yes both avoid resampling hexagonal data onto a square grid.
> 
> The difference is the mechanism: HexagDLy (https://arxiv.org/pdf/1903.01814) uses an offset-column addressing where the hexagonal neighbourhood is exactly representable in 2D, and decomposes the hex kernel into rectangular sub-kernels, so standard convolutions do the work. This PR is just the mapper producing that addressing, with the neighbour topology verified exactly.
> 
> The practical gain is memory: nothing extra gets materialised at each layer, so memory barely grows when you use larger kernels, which can allow a bigger batch size. The cost is speed: HexagDLy is slower per step.
> 
> I'm also working on the hls4ml support, and we're currently running some tests on it for SST-1M. (For now I consider that only the io_stream mode for Conv2d and MaxPool2d is working; I still need to think about how to do it in 3D, though if I'm not wrong, even the standard Conv3D layer isn't supported in hls4ml yet.)
> 
> Another advantage is the new share_neighbors parameter on Conv2d and Conv3d, which could help a model learn symmetrical/diagonal/ring patterns while reducing the number of weights and keeping the same performance, this could help for an FPGA implementation, even though the number of operations stays the same.
> 
> As it's the direction we took for SST-1M, I was wondering if it could be useful for CTA.
> 
> Not proposed as a replacement for IndexedConv, just another no-resampling route.

Excellent, thank you very much for the thorough explanation and background details !

